PR #5, opened 2019-08-13
BEAM

karanpatel863:
Beam is a next generation scalable, confidential cryptocurrency based on an elegant and innovative Mimblewimble protocol.

Things that make BEAM special include:

Users have complete control over privacy - a user decides which information will be available and to which parties, having complete control over his personal data in accordance to his will and applicable laws.
Confidentiality without penalty - in BEAM confidential transactions do not cause bloating of the blockchain, avoiding excessive computational overhead or penalty on performance or scalability while completely concealing the transaction value.
No trusted setup required.
Blocks are mined using Equihash Proof-of-Work algorithm.
Limited emission using periodic halving.
No addresses are stored in the blockchain - no information whatsoever about either the sender or the receiver of a transaction is stored in the blockchain.
Superior scalability through compact blockchain size using the “cut-through” feature of Mimblewimble.
No premine. No ICO. Backed by a treasury, emitted from every block during the first five years.
Implemented from scratch in C++.

BEAM has set out to extend the feature set of Mimblewimble in a number of ways. BEAM supports setting an explicit incubation period on a UTXO, which limits its ability to be spent to a specific number of blocks after its creation. This is different to a timelock, which prevents a transaction from being added to a block before a certain time. BEAM also supports the traditional timelock feature, but includes the ability to also specify an upper time limit, after which the transaction can no longer be included in a block. This feature means that a party can be sure that if a transaction is not included in a block on the main blockchain after a certain time, it will never appear.

Another unique feature of BEAM is the implementation of an auditable wallet. For a business to operate in a given regulatory environment, it will need to demonstrate its compliance to the relevant authorities. BEAM has proposed a wallet designed for compliant businesses, which generates additional public/private key pairs specifically for audit purposes. These signatures are used to tag transactions so that only the auditing authority that is given the public key can identify those transactions on the blockchain, but cannot create transactions with this tag. This allows a business to provide visibility of its transactions to a given authority without compromising its privacy to the public.

BEAM's implementation of Dandelion improves privacy by adding decoy transaction outputs at the stem phase. Each such output has a value of zero, but is indistinguishable from regular outputs. At a later stage (a randomly calculated number of blocks for each output), the UTXOs are added as inputs to new transactions, thus spending them and removing them from the blockchain.

BEAM has also proposed another feature aimed at keeping the blockchain as compact as possible. In Mimblewimble, as transactions are added, cut-through is performed, which eliminates all intermediary transaction commitments. However, the transaction kernels for every transaction are never removed. BEAM has proposed a scheme to reuse these transaction kernels to validate subsequent transactions. In order to consume the existing kernels without compromising the transaction irreversibility principle, BEAM proposes that a multiplier be applied to an old kernel by the same user who has visibility of the old kernel, and that this be used in a new transaction. In order to incentivize transactions to be built in this way, BEAM includes a fee refund model for these types of transactions. This feature will not be part of the initial release.

When constructing a valid Mimblewimble transaction, the parties involved need to collaborate in order to choose blinding factors that balance. This interactive negotiation requires a number of steps and it implies that the parties need to be in communication to finalize the transaction. Grin facilitates this process by the two parties connecting directly to one another using a socket-based channel for a "real-time" session. This means that both parties need to be online simultaneously. BEAM has implemented a Secure Bulletin Board System (SBBS) that is run on BEAM full-nodes to allow for asynchronous negotiation of transactions.

Requiring the interactive participation of both parties in constructing a transaction can be a point of friction in using a Mimblewimble blockchain. In addition to the secure BBS communication channel, BEAM also plans to support one-sided transactions where the payee in a transaction who expects to be paid a certain amount can construct their half of the transaction and send this half-constructed transaction to the payer. The payer can then finish constructing the transaction and publish it to the blockchain. Under the normal Mimblewimble system this is not possible, because it would involve revealing your blinding factor to the counterparty. BEAM solves this problem by using a process it calls kernel fusion, whereby a kernel can include a reference to another kernel so that it is only valid if both kernels are present in the transaction. In this way, the payee can build their half of the transaction with a secret blinding factor and a kernel that compensates for their blinding factor, which must be included when the payer completes the transaction.

### DIFF
--- a/data/BEAM
+++ b/data/BEAM
@@ -1,0 +1,24 @@
+# Ecosystem Level Information
+title = "BEAM"
+
+# Github Organizations
+# This is a list of links to associated github organizations.
+github_organizations = ["https://github.com/BeamMW/beam", "https://github.com/BeamMW/android-wallet", "https://github.com/BeamMW/beam-docs", "https://github.com/BeamMW/ios-wallet"]
+
+# Repositories
+# These are structs including a url and tags for a git repository.  These URLS do not have to be on github.
+[[repo]]
+url = "https://github.com/BeamMW/beam"
+tags = [ "Beam MimbleWimble"]
+
+[[repo]]
+url = "https://github.com/BeamMW/beam-docs"
+tags = [ "Beam Documentation"]
+
+[[repo]]
+url = "https://github.com/BeamMW/android-wallet"
+tags = [ "Android Wallet"]
+
+[[repo]]
+url = "https://github.com/BeamMW/ios-wallet"
+tags = [ "iOS Wallet"]


jubos:
Thank you for the PR, @karanpatel863.  The beam ecosystem is already added here: https://github.com/electric-capital/crypto-ecosystems/blob/master/data/ecosystems/b/beam.toml.  Please add any new repos that missing if you find them.

karanpatel863:
thanks jubos. will keep adding new repos as they are being developed ^^

PR #8, opened 2019-08-22
adds Aleph Zero ecosystem

mike1729:


data/ecosystems/a/aleph-zero.toml
```diff
@@ -0,0 +1,16 @@
+# Ecosystem Level Information
+title = "Aleph Zero"
+
+# Sub Ecosystems
+# These are the titles of other ecosystems in different .toml files in the /data/ecosystems directory
+sub_ecosystems = []
+
+# Github Organizations
+# This is a list of links to associated github organizations.
+github_organizations = ["https://github.com/alephzerofoundation"]
```
jubos: The alephzerofoundation is actually a user, not an organization.  You can just use an empty array here for now.  I suggest moving alephzerofoundation to be an organization since it is a better container for repositories and adding contributors over time. 
```suggestion
github_organizations = []
```

mike1729: Thanks for the advice, we have created an organization aleph-zero-foundation and I've updated the toml file.

PR #9, opened 2019-08-23
Add Chainstack

akegaviar:
Add Chainstack -- a blockchain deployment platform similar to Infura
but with multi-protocol support


jubos:
Thank you @akegaviar for the PR!  Merged.

PR #10, opened 2019-08-24
analysis example in python

benjyz:
an example of analysis in python, not you want to merge this, but leaving the PR for reference in any case. do you do the data analysis in rust?

### DIFF
--- a/analysis/repo_example.py
+++ b/analysis/repo_example.py
@@ -1,3 +1,7 @@
+"""
+basic example to analyse the repos
+"""
+
 import toml
 import os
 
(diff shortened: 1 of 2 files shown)


benjyz:
sample output

```
analysing  /Users/ben/github/crypto-ecosystems/data/ecosystems
orgs  1624
repos  27489
repos with bitcoin in its name 312
repos with trading in its name 18
```

jubos:
Thanks for the pull request.  Like you mentioned, I will keep analysis separate to keep the focus on the data in this core repository.  I added a subset of your example to the wiki and will keep this PR around for posterity if people search for an analysis example.

PR #12, opened 2019-08-30
Adding Liquality

h4v0kr:
Atomic Swap between BTC, DAI, and ETH with minimal risk and fees, while maintaining control over your private keys


jubos:
Thank you for the PR!  Merged.

PR #23, opened 2020-06-01
Remove copied repos

theoturner:
Removal of copied repos for Electron Cash (clones of FreeBSD and the Android NDK with no changes). These add several hundred thousand commits to index but are in no way Bitcoin Cash repos.

### DIFF
--- a/data/ecosystems/b/bitcoin-cash.toml
+++ b/data/ecosystems/b/bitcoin-cash.toml
@@ -107,12 +107,6 @@ url = "https://github.com/dsmurrell/awesome-bitcoin-cash"
 url = "https://github.com/Ekliptor/cashp"
 
 [[repo]]
-url = "https://github.com/Electron-Cash/android-platform-ndk"
-
-[[repo]]
-url = "https://github.com/Electron-Cash/android-vendor-freebsd"
-
-[[repo]]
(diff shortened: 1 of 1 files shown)


jubos:
Hi @theoturner, thank you for the PR.  We actually keep the forks in there in case we detect that the teams are contributing their own commits on top of the forked repositories.  In many cases they don't add anything, but in cases like zcash or litecoin, they add a lot of their own commits on the original forked code base.

theoturner:
This is not a matter of forks in general, it is considering a case by case basis. There is no association between an unmodified FreeBSD and Bitcoin Cash, it only adds several minutes of processing time for no additional data. If Bitcoin Cahs were to modify the repo, it should be added at that point.

jubos:
Hard to know apriori if the Bitcoin Cash team has not made changes to that repo -- hence its addition.  It's also easier to automate since repos added to the github organization are picked up.  Usually someone forks a repo if they hope to make changes on it at some point in the future, but I agree often it might just sit there with no downstream changes.

In your own processes you could detect if it is a forked repo and just ignore it if you don't want to process it?

theoturner:
For those wanting a process, github displays commits ahead/behind above the code. If the repo is no commits ahead, then there are no changes.

PR #27, opened 2020-09-17
use the actual Astake repo instead of an out of date one

elliotolds:


jubos:
Thanks for the PR, merged!

PR #28, opened 2020-09-30
Update ethereum.toml

MoonieMama:
Adding Pooltogether to the Ethereum ecosystem tool


jubos:
Thanks for the PR @MoonieMama.  If you add a sub ecosystem you need to create a corresponding toml for it otherwise it won't pass the validation checks.  You can just add the repos with no sub ecosystem for simplicity.

jubos:
@MoonieMama, I handled adding PoolTogether to the Ethereum ecosystem separately.  Thank you for the PR.

PR #29, opened 2020-11-08
Add various projects to the Solana and Serum ecosystem pages

rmshea:
A few projects related to Solana and Serum were missing from their pertaining ecosystem files. I went ahead and added the missing projects to each.

An open question: Solana has multiple partnerships announced [here](https://solana.com/ecosystem) that don't have public GitHub links. I noticed that only git repositories are being added — should I leave these partnerships and engagements out or add in the relevant announcements on other domains?

Many thanks.

### DIFF
--- a/data/ecosystems/p/project-serum.toml
+++ b/data/ecosystems/p/project-serum.toml
@@ -50,3 +50,9 @@ url = "https://github.com/project-serum/swap-ts"
 
 [[repo]]
 url = "https://github.com/project-serum/validators"
+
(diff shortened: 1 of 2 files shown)


jubos:
Thanks for the PR @rmshea.  As for partnerships, feel free to do new PRs when they open up their repos, and I can also watch for them as well.  

You can also add their repos to respective ecosystems as well. This often happens in an automated fashion by looking at their github organizations.

PR #34, opened 2020-12-02
add Tezos Domains to the Tezos repo list

mbodecek:


jubos (changes requested):
This is fine to leave this fork in place in case development happens there.  We have a lot of logic to dedup commits across forks.

jubos:
Thanks for the PR, merged

PR #37, opened 2020-12-05
Add Stafi related repositories

Tore19:
Add more repositories for Stafi.

Many thanks.

### DIFF
--- a/data/ecosystems/s/stafi.toml
+++ b/data/ecosystems/s/stafi.toml
@@ -3,7 +3,7 @@ title = "Stafi"
 
 sub_ecosystems = []
 
-github_organizations = ["https://github.com/stafiprotocol"]
+github_organizations = []
 
 # Repositories
 [[repo]]
--- a/data/ecosystems/s/stafi.toml
+++ b/data/ecosystems/s/stafi.toml
@@ -1,0 +1,34 @@
+# Ecosystem Level Information
+title = "Stafi"
+
+sub_ecosystems = []
+
+github_organizations = ["https://github.com/stafiprotocol"]
+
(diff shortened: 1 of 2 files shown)


jubos:
Thanks for the PR.  If you want to add Stafi as a subecosystem you need a make a separate toml file for it and then reference it in the polkadot.toml.

jubos:
If you have the rust toolchain installed, you can also run "make validate" to ensure the dataset is valid before uploading your changes.

Tore19:
@jubos Hi, Just changed as you said. If there is any problem, please let me know. Thanks!

jubos:
@Tore19, thanks for the changes.  One other fix is that stafiprotocol is actually a github user and not an organization.  You can either remove it or upgrade stafiprotocol to an org on github.

data/ecosystems/s/stafi.toml
```diff
@@ -0,0 +1,34 @@
+# Ecosystem Level Information
+title = "Stafi"
+
+sub_ecosystems = []
+
+github_organizations = ["https://github.com/stafiprotocol"]
```
jubos, line 6: This is just a github user as opposed to an org, so you can make this an empty array for now.

jubos, line 6: This is just a github user as opposed to an org, so you can make this an empty array for now.

jubos (changes requested):
Just need to tweak the github_organizations array.

data/ecosystems/s/stafi.toml
Tore19: Done! Thanks!

PR #40, opened 2020-12-08
Add assorted neo repos

EdgeDLT:


jubos (approved):
Looks great, thanks for all these new repos.

PR #45, opened 2020-12-15
Update cosmos.toml

ebuchman:
Added Agoric, Althea, Binance, Cyber, Enigma, FOAM, LazyLedger, Nomic, and Nym to Cosmos

### DIFF
--- a/data/ecosystems/c/cosmos.toml
+++ b/data/ecosystems/c/cosmos.toml
@@ -13,7 +13,7 @@ sub_ecosystems = [
   "Interchain",
   "IRIS Network",
   "Kava Labs",
-  "LazyLedger",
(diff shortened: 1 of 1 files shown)


jubos:
Thanks for the PR.  This is very helpful.  Do you have a sense for when something graduates from using Tendermint to becoming part of Cosmos?

ebuchman:
That's a good question. Generally speaking, any project that uses the Cosmos-SDK for building its app on Tendermint (most of the ones included here), but also projects that identify as part of the larger Cosmos ecosystem through intention to integrate with IBC, even if they're not built using the Cosmos-SDK (eg. Nomic).

jubos:
Thanks, merged!

PR #49, opened 2020-12-18
Add 3box as a subecosystem of IPFS 

andrew:


jubos (changes requested):
I think it needs to be "3Box" to pass validation.

PR #51, opened 2021-01-19
Update cosmos.toml

okwme:
Peggy repo was renamed recently, not sure how it effects historical tracking


jubos:
Thanks for the PR.

PR #54, opened 2021-01-28
Update monero.toml

Agorist-Action:
Adding my Monero C# RPC Library, which was added to the Monero Ecosystem [here](https://github.com/monero-ecosystem/csharp-monero-rpc-client).


jubos:
Thanks for the PR.

PR #63, opened 2021-04-17
Update celo.toml for Project SmartLoan

Asvoria:
Update celo.toml for Project SmartLoan.


jubos (changes requested):
This looks like a duplicate with #64.  Can you merge these two and rebase on master?

PR #70, opened 2021-05-17
celo.toml updates again

aslawson:


jubos:
Thanks for this.  It looks like it might require a rebase to make it build.

jubos (changes requested):
Looks like this needs a rebase

PR #79, opened 2021-09-17
Update list of Dfinity (internet computer) projects

bitdivine:
# Motivation
Someone drew my attention to crypto-ecosystems.  I too have a list of Dfinity (Internet Computer) projects, with everything from personal projects (e.g. https://github.com/andyzal/icstatica) and a canister that itself gets adoption stats (https://github.com/dprats/icp-ecosystem) to full blown projects like https://github.com/getimpactnow/getimpactnow (which has a [website](https://getimpactnow.org/) and the team is working on an app).

# Changes
This PR adds the projects I was able to find to your list.  Thank you for those on your list that I didn't have yet.

# Notes
Are there particular criteria you need a project to fulfil to be listed?  This list is suitable for general adoption but if you need only projects with thousands of users this would have to be filtered quite a bit.  Maybe we could also add project URLs, or we can try to get the live number of users for these canisters.  What do you think?  Getting live stats would require some work but could be done.  Maybe not soon, as I have a new family member about to land, but the data is all available to do so, I think, once we have domain names.

### DIFF
--- a/data/ecosystems/d/dfinity.toml
+++ b/data/ecosystems/d/dfinity.toml
@@ -4,14 +4,371 @@ title = "DFINITY"
 sub_ecosystems = []
 
 github_organizations = [
+  "https://github.com/1082-xyz",
+  "https://github.com/109km",
+  "https://github.com/3cL1p5e7",
+  "https://github.com/55foundry",
+  "https://github.com/aitikgupta",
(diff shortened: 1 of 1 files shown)


jubos:
Merged this.  Removed the github users that were miscategorized as orgs and some other non-dfinity orgs.

PR #126, opened 2021-10-21
Update celo.toml

steffenkrogmann:
Adding Kukuza


jubos:
Thanks, merged this in.

PR #154, opened 2021-11-03
Update celo.toml to include Trans Pharma

moreplatesmoredates:


jubos:
Thank you, merged this.

PR #171, opened 2021-12-15
Add AshSwap to Elrond ecosystem

phanletrunghieu:
Our first public product, AshSwap - a stable-swap exchange, is built on Elrond.
We already have testnet version up and running at https://testnet.ashswap.io.
The token faucet at: https://faucet.ashswap.io.
Code can be found at https://github.com/ashswap.


jubos:
Thanks for this!

PR #178, opened 2021-12-20
Update the Internet Computer ecosystem level information

ililic:
This repo had a stale DFINITY (IOU) repository and the Dfinity ecosystem repository should be under the "Internet Computer" brand.

We've also added a number of ecosystem projects that weren't included in the existing `toml` file.


jubos:
Thanks for fixing this up.

PR #185, opened 2022-01-06
Update maker.toml

Jevans13:
Maker toml file updated to reflect the addition of Core Unit specific repo's added since the move to full DAO operation.


jubos:
Merged, appreciate this very comprehensive update.

PR #192, opened 2022-01-08
rename Evmos, add Cronos to cosmos ecosystem

fedekunze:


jubos:
Thank you for these changes!

PR #203, opened 2022-04-30
Adding more repos

muttoni:
Adding more repos, especially those that integrate with FCL.js, the Flow Client Library (i.e. any app building on Flow)

### DIFF
--- a/data/ecosystems/f/flow.toml
+++ b/data/ecosystems/f/flow.toml
@@ -489,16 +489,7 @@ url = "https://github.com/yellowBirdy/flow-worksop-basic"
 url = "https://github.com/yellowBirdy/monoswap"
 
 [[repo]]
-url = "https://github.com/DappStarter/flowery-nenetir-mQbm"
-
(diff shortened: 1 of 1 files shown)


jubos (changes requested):
Thanks for this update.  Can you remove the DappStarter templates? These are mostly just boilerplate repos.

muttoni:
Done!

jubos:
Thanks for removing those.  I think this PR still needs a bit of filtering.  Are you running an automated process to suggest these repos?  For instance a few are more general and not specific to flow:

https://github.com/RobEnoch/Tatum-js
https://github.com/magiclabs/magic-js
https://github.com/David405/algorand-lending-protocol

In general for tooling that supports multiple ecosystems, those are better suggestions for the General ecosystem. 

Can you do another pass?

muttoni:
Hey @jubos, are we good to merge this now?

jubos:
Thanks merged this!

PR #205, opened 2022-06-10
Add webln.toml

MoritzKa:


jubos (changes requested):
Thank yoy. This probably is best added as a direct repository to the Lightning ecosystem.

jubos:
This is already covered in the Lightning ecosystem, so closing this out.

PR #213, opened 2022-08-04
Pull repo stats

forgetso:
The following modifications to this repo can be used to pull stats on each of the repositories. This helps to identify if they are still actively developed or even offline (status=`404`).

Simply build and leave to run with this command:

```bash
./target/debug/crypto-ecosystems update ./data
```

I did experiment with scraping asynchronously but quickly hit 429 Too Many Requests from GitHub.


jubos:
Cool idea!  This is a bit beyond the scope of this taxonomy since the validate call should be synchronous and not require any kind of network/IO activity to succeed.  It needs to be runnable by the Github actions.  We are also keeping around old repository links since they did exist at some point in time to keep the whole history.  We may change this at some point. 

Creating another repo with this as a functionality could be useful for people contributing to crypto-ecosystems if they wanted to validate particular repos.

PR #226, opened 2022-09-02
update with web3.storage and nft.storage repos

dchoi27:


jubos:
Merged this, thank you for the cleanup.

PR #228, opened 2022-09-04
Add Kamu and Open Data Fabric to the taxonomy

sergiimk:


jubos:
Thanks for these ecosystems, merged!

PR #229, opened 2022-09-04
Update telos.toml

nathanduft44:
Please add Telos data, we're growing quickly.


jubos (changes requested):
For these sub ecosystems, each one needs its own toml file if it doesn't exist already.

PR #235, opened 2022-09-16
Added repositories to solana.toml

notEduardo:


jubos:
Thanks for monster PR!  We are running these through review and will circle back shortly.

0xnirmal:
anything I can do to help @jubos? would love to see this get merged in!

jubos (changes requested):
@notEduardo, thanks for this diff, but we can't merge it as is.  There are a lot of duplicates that are already reflected in the sub ecosystems of Solana.  It also casts too wide of a net and pulls in a lot of repos that are not dedicated to Solana like ankr and badger to name a few.

As a next step, if you can scope down the diff into smaller reviewable diffs, it will be much easier to confirm and get them in.  We have also reviewed a bunch of Solana repos internally and will make an update to Solana.toml that should make the job easier.

jubos:
Going to close this out for now.  ff8b5e619faa8258f9e2598593958141f1bf549b added a bunch of new Solana repos + sub ecosystems.  Please rebase against this when you have a moment and reopen the PR with any missing repos.

PR #241, opened 2022-10-03
Update polkadot.toml

2075:


jubos:
Fixed the gamedaoco dup and merged this.  Play Zero and GameDao are already sub ecosystems of Polkadot, so are fully covered.

PR #246, opened 2022-10-04
Update InvArch and add it to Kusama

arrudagates:
Self-explanatory title, this PR updates some repositories from invarch.toml and also adds it as a sub_ecosystem in Kusama too.


spazcoin:
sorry I missed InvArch when adding last night 😞 
I searched for Youdle, though. (found nothing)

jubos:
Merged this.  Thanks!  

I kept some of the repos you removed just to maintain the history there.

PR #255, opened 2022-10-11
Programmatically Updating all Broken Links

neosavvy:
We have written a program that leverages Electric Capital's index to review all histories of Web3 / crypto (or any git repo, for that matter) for complexity.

The first step in this process is to clone that repository. We noticed that a huge subset of the repositories in this index is no longer accessible.

This is a draft PR to share this information back with those who depend on it. It would be great to have a conversation to see what else we could do in terms of base-level foundation data in web3/crypto.

Please let me know if you'd like other things done programmatically at scale.

I'm reachable at adam@foliofficient.com


neosavvy:
Quickly rebased this morning to make sure this is up to date. I suspect you all may want to suggest a tag that would be a good gravestone for posterity's sake. 

I chose "404" as these URLs are all dead when I sync them down to analyze commits. Let me know if you think we should use something else, or if removing them entirely would be the preference. 

After talking to @jubos on twitter the other day, I presumed that removing them would result in folks never looking back in git logs. Leaving a "grave" marker at least lets people know that at some point there was a product/project/repo at a point in time. 

This also shows community churn for those who are calculating such metrics.

jubos:
Thanks for this idea, @neosavvy.  I am just following up here.  I think the way forward is to make a separate metadata attribute "archived" in the toml representation.  

I am going to work on that more holistically in #268 and will keep you posted.

PR #267, opened 2022-10-26
added new repos from Elrond Community

schimih:
Updated the repo list for Elrond.


jubos:
Looks good, thanks for these new ecosystems.  Merged!

PR #272, opened 2022-11-01
Add Gafi Network ecosystem - Polkadot

mutobui:


jubos:
Merged just the first one, since you don't need to add the repos to polkadot directly since they are captured in the sub ecosystem.

Thanks!

PR #274, opened 2022-11-06
Added repositories to Solana.toml

notEduardo:
previous conversation 
https://github.com/electric-capital/crypto-ecosystems/pull/235


jubos:
This is still pulling in way too many non Solana repos and duplicates, so cannot review as is.  Many of these repos are already reflected in the Solana sub ecosystems. 

If possible, can you please do what the moonbeam contributors have been doing and make smaller digestible PRs that can be easily reviewed?

PR #278, opened 2022-11-07
Create hashport.toml

bmgentile:
Creation of hashport.toml for the Hedera ecosystem


jubos (changes requested):
Looks good, please remove the github_org since it will pull in extra repos that aren't related to hashport.

bmgentile:
Thanks @jubos I've gone ahead and updated the PR.

PR #299, opened 2022-11-08
update letters U, V, W ecosystems

themacexpert:
add / update letters U, V, W ecosystems


jubos:
Fixed conflicts and merged

PR #300, opened 2022-11-08
update letters XYZ ecosystems

themacexpert:
add / update letters XYZ ecosystems


jubos:
Fixed conflicts and merged

PR #302, opened 2022-11-16
add moonriver subecosystems

themacexpert:
Add Moonriver subecosystems


themacexpert:
Hi @jubos,  everything look good here?

jubos:
@themacexpert, it's looking closer.  It contains mostly sub ecosystems that aren't primary to Moonriver here like API3, Ceramic, DODO, Gelato, Guild, etc.

We haven't exposed "secondary" ecosystems yet in the taxonomy.  We hope to do that sometime next year.  

A few good rules of thumb: 
1/ if the governance and token originates on the network, it is likely a sub ecosystem.
2/ if you go to their web page and it **only** mentions that ecosystem, it is likely a sub ecosystem 

Dot Finance, for instance, fits the bill of a sub ecosystem to Polkadot since all of their copy on their website and docs is about Polkadot.  

In this case, Moonsama is probably a good example of Moonriver specific sub ecosystem.

themacexpert:
thanks @jubos, should be good now I think

jubos:
Awesome, thank you!

PR #303, opened 2022-11-16
add Moonbeam subecosystems

themacexpert:
add Moonbeam subecosystems


themacexpert:
Should be good to go also @jubos

jubos (approved):

PR #304, opened 2022-11-16
add and update various projects

themacexpert:
add and update various projects


jubos:
Thanks for these updates!

PR #319, opened 2022-11-29
urbit: add ecosystems, organizations, repos

jalehman:
The existing file is missing repositories from organizations in our ecosystem — this PR adds many of them.


jubos:
This is fine for now.  Over time, let's move these github orgs into their own sub ecosystems if they are indeed separate.

PR #320, opened 2022-11-29
2022 polkadot.toml file update

olliecorbisiero:
Updated polkadot.toml file for 2022 Electric Capital Report. 

Incorporated updates to Sub Ecosystems, Github Organizations & Github Repos. 

The update was done programmatically hence why 2,097 additions and 1,054 deletions.

Thank you!

### DIFF
--- a/data/ecosystems/p/polkadot.toml
+++ b/data/ecosystems/p/polkadot.toml
@@ -2,243 +2,388 @@
 title = "Polkadot"
 
 sub_ecosystems = [
-  "5ire",
-  "Acala",
-  "Acurast",
-  "Adoria Soft",
-  "Airalab",
-  "AirGap",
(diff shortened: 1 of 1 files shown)


data/ecosystems/p/polkadot.toml
```diff
@@ -2,244 +2,389 @@
 title = "Polkadot"
 
 sub_ecosystems = [
```
jubos, line 4: Many of these subecosystems need to be added in their own toml files for the build to succeed.  Can you please add those in separate PRs and then do one PR to add them all to Polkadot to make for easier review?

jubos (changes requested):

jubos:
Looks like this is still failing.  Shall we abandon this in favor of #325?

PR #322, opened 2022-12-01
Update to Hedera.toml

bmgentile:
Updating hedera.toml to reference new projects in the Hedera ecosystem + contributor repos.


bmgentile:
Additionally, when does the official 2022 report become available / when do I need to have all updates completed for them to be included?

PR #305, opened 2022-11-17
Adding Cosmos/Umee

robert-zaremba:
Redoing the following PR, which was not merged (apparently some error was there)

PR #330, opened 2022-12-09
Updating Umee and adding Ojo

robert-zaremba:
Follow up to #305

#305 was not merged, instead Umee was manually added to the ecosystem list. However this action:
+ doesn't include Peggo

This PR:
+ adds Ojo repository
+adds Peggo
+ explicitly adds Umee chain repository (otherwise, I don't see how Umee will be picked up just based on the ecosystem name). Umee is an app chain, not a fork of Cosmos SDK.

Note, we also maintain our own forks of Cosmos SDK and Gravity Bridge, so not sure if these should be included as well.


jubos:
Sorry if it's not clear, but everything in the Umee subecosystem will implicitly be pulled up into the parent, which is Cosmos.  That way we don't need to duplicate the repos in umee also in the cosmos.toml.

As for ojo-network, it should likely be its own ecosystem, which could be a sub ecosystem of Umee if it is primarily built for Umee.  If it is more of a general purpose app-chain for Cosmos, I would make Ojo a sub ecosystem of Cosmos instead.

robert-zaremba:
Oh, I see. This is by importing `data/ecosystems/u/umee.toml` into Cosmos? So I guess we are good for the umee org, and we don't need to update anything?

PR #337, opened 2022-12-19
december update for MultiversX

schimih:
December update for MultiversX.


jubos:
Manually merged this one into the multiversX toml file.  Thank you for all of these repos!

Issue #268, opened 2022-10-29
Add "archived" boolean attribute on old repositories


jubos:
Some repositories get removed or archived, so add a special attribute on those entries in the toml to reflect that.

neosavvy:
Thanks for having a look, let me know if you want me to submit a Python CLI that helps you check everything in the repo and write your tag/attribute back. Happy to add it to my list as it'll benefit me and anyone using this index.

jubos:
This is called "missing" now since archived implies something different on github.

PR #343, opened 2022-12-21
remove old / moved / hidden Axelar repositories

StephenFluin:
This PR removes repositories that are no longer public or have been moved within the `axelarnetwork` GitHub organization.

**NOTE:** I'm not sure how the electric-capital system works. If these need to be here to pull data from earlier in 2022, feel free to reject this PR as these USED to be valid repositories.


jubos:
Thanks Stephen, we keep even the old ones around for posterity.  We are working on #268 to denote archived repos in the future, so that will help out here.

PR #345, opened 2022-12-21
Update tezos.toml

upsorber:
Add "Upsorber" to sub-ecosystem.


jubos:
Upsorber doesn't look to have any open source repos beyond a gist.  Let's reopen this when there are some github repos.

PR #346, opened 2022-12-21
Update tezos.toml

jordan-tezos:
Add "SmartPy" to sub_ecosystems and SmartPy's public repository.


jubos:
Moved smartpy to a repo instead of a github org and manually merged it.

PR #350, opened 2022-12-22
Create sns.toml

mistertoucan:


data/ecosystems/s/sns.toml
```diff
@@ -0,0 +1,24 @@
...
+[[repo]]
+url = "https://github.com/snsdomains/objects.json"
+
+# Private Repositories (will be made public)
+
+[[repo]]
+url = "https://github.com/snsdomains/domains"
+
+[[repo]]
+url = "https://github.com/snsdomains/web"
```
jubos, line 24: Let's remove these for now since we only track open source repos in the taxonomy at the moment.  The github_organization being specified above will automatically add them in the future when they are open sourced.

jubos (changes requested):

data/ecosystems/s/sns.toml
mistertoucan: Sounds good. I just removed them. Please check now

PR #352, opened 2022-12-22
add Oxhead Alpha as a Tezos sub-ecosystem

nicolasochem:


data/ecosystems/t/tezos.toml
```diff
@@ -36,6 +37,7 @@ github_organizations = [
   "https://github.com/cryptiumlabs",
   "https://github.com/hicetnunc2000",
   "https://github.com/kukai-wallet",
+  "https://github.com/oxheadalpha",
```
jubos, line 40: This gets pulled in automatically because 0xhead Alpha is a sub eco of tezos.  Can you please remove it and then it is good to go!

jubos (changes requested):

jubos:
Thank you for this!

PR #353, opened 2022-12-22
Update lightning.toml

lucasdcf:
Added a few repos!


jubos:
Thanks for these repos!  Merged.

PR #360, opened 2022-12-22
feat: add Tezos ecosystem repos

RichAyotte:


jubos:
Thanks @RichAyotte.  I added baktz as a repo to tezos and skipped Exodus because it is more general that just Tezos.  We can add Baktz as a sub ecosystem later if you think it's appropriate.

PR #378, opened 2023-01-06
Add Stacks Degens

BowTiedDeployer:


data/ecosystems/s/stacks.toml
```diff
@@ -10,6 +10,7 @@ sub_ecosystems = [
   "hiro-systems",
   "jolocom",
   "lnstxbridge",
+  "Stacks Degens",
```
jubos, line 13: You need to create a sub ecosystem toml file if you want this to pass the build correctly.

jubos (changes requested):

wileyj:
this can be closed per #381

data/ecosystems/s/stacks.toml
BowTiedDeployer: Was updated in the #381 PR.

PR #387, opened 2023-01-07
splits gitcoin from gitcoin => 1/2

owocki:
(created upon request from EC/gitcoindao)


jubos:
Does Gitcoin 2 have a different name we can use to denote a separate ecosystem ala Gitcoin Classic and Gitcoin?

owocki:
@jubos Centralized Gitcoin vs GitcoinDAO  would work.

PR #391, opened 2023-01-07
Update chia.toml - add smart farm

scrutinously:
While not directly tied to the crypto, this repo is for automating the monitoring of hard drive health and temperatures, which is useful for farming Chia.


jubos:
Looks reasonable to add.  Thanks!

PR #398, opened 2023-01-08
Update chia.toml

MarvinQuevedo:
Chia ozone repository  and BLS Flutter binding for use Native cpp libraries in Flutter(Android and iOS)


jubos:
Thanks!

PR #412, opened 2023-01-10
Update chia.toml

farmerdiamonds:
Added 3 more repositories and changed position because of the alphabetic order


jubos:
Thanks!

PR #413, opened 2023-01-10
rename organization to multiversx

schimih:
We migrated all our repos to the new organization `https://github.com/multiversx`


jubos:
Thanks for the cleanup @schimih!

PR #422, opened 2023-01-17
remove non-bitcoin projects

jlopp:
The projects either no longer exist or are solely dedicated to an ecosystem other than BTC


jubos:
Thank you for cleaning these up!

PR #427, opened 2023-01-18
Update Akash Network

gosuri:
We recently migrated Akash related source code to a dedicated [organization](https://github.com/akash-network).


jubos:
Thanks @gosuri!

PR #431, opened 2023-01-18
Adding Algorand for Visual Studio

FrankSzendzielarz:


jubos (changes requested):
This looks like it is just a single repository for now, so I would just add it to the Algorand ecosystem.

A good rule of thumb for whether something should be its own ecosystem is if it has its own website, community (discord, telegram), twitter account, and github organization.

eherrerosj:
I've added the repo URL manually, closing PR, thanks for contributing @FrankSzendzielarz

PR #435, opened 2023-01-19
update iotex.toml

GiupiDeLuca:
added machinefi and relative repos, and updated iotexproject repos


jubos:
Looks good to me, thanks!

PR #440, opened 2023-01-21
[f-Functionland] Removed outdated repos and added tags

ehsan6sha:


jubos:
Thank you for the cleanup!

PR #441, opened 2023-01-21
More bitcoin repos

jlopp:


data/ecosystems/b/bitcoin.toml
```diff
@@ -645,19 +672,347 @@ url = "https://github.com/bitpay/bitcore-node"
...
+url = "https://github.com/BlockchainCommons/bc-lifehash-python"
+
+[[repo]]
+url = "https://github.com/BlockchainCommons/bc-shamir"
+
+[[repo]]
+url = "https://github.com/BlockchainCommons/bc-sskr"
+
+[[repo]]
+url = "https://github.com/BlockchainCommons/bc-ur"
```
jubos, line 735: There is a Blockchain Commons ecosystem already.  Do you think it's appropriate to make that a direct sub ecosystem of Bitcoin?

jubos (changes requested):

data/ecosystems/b/bitcoin.toml
jlopp: Ah, I missed that. Yes, I'll amend to make it a sub system. While blockchain commons does have a lot of generalized security work that could be applied to management of any cryptographic keys, all of their protocol specific projects are only Bitcoin related.

jubos:
Looks good, thanks!

PR #450, opened 2023-01-25
Update polymesh.toml and polymath.toml

polymath-eric:
Generally "Polymath" refers to the deprecated ERC 20 ethereum based financial securities smart contracts

"Polymesh" is the Substrate based replacement as a purpose built chain


jubos:
Thank you for the clarification and clean up here.

PR #456, opened 2023-02-04
cardano updates 1

Blugh:
minor updates for cardano ecosystems


jubos:
Thanks for these additional ecosystems.

PR #457, opened 2023-02-05
Update Cardano.toml to include Aiken repository url

teddyjfpender:
Added missing Aiken repository URL in Cardano


jubos:
Looks good, thank you.

PR #458, opened 2023-02-06
Adding Q Protocol to taxonomy

sameersg:
Would like to add Q Protocol from q.org


jubos (changes requested):
Thank you for the PR.

These tags are too specific to Q-Testnet.  Good examples of tags are "Documentation", "Protocol", "Wallet", etc.

"Monitoring" is a reasonable one to leave in.

sameersg:
Sure let me update the tags

sameersg:
thanks

PR #471, opened 2023-03-03
Additional Flow repos

muttoni:
Added new Flow repos and removed a couple incorrect ones as I QA'd the list

### DIFF
--- a/data/ecosystems/f/flow.toml
+++ b/data/ecosystems/f/flow.toml
@@ -5866,8 +5866,6 @@ url = "https://github.com/script-money/flowTokenBatchSend"
 [[repo]]
 url = "https://github.com/dangdennis/flow-runner"
 
-----HACKATHON---
-
 [[repo]]
 url = "https://github.com/LyricLabs/render-fe"
 
--- a/data/ecosystems/f/flow.toml
+++ b/data/ecosystems/f/flow.toml
@@ -2592,9 +2592,6 @@ url = "https://github.com/LearnWeb3DAO/Flow-Track"
 url = "https://github.com/leavs-crypto/leavs"
 
 [[repo]]
-url = "https://github.com/legendaryphoenixicarus/Solana-mint-NFT"
-
-[[repo]]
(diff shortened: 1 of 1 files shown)


data/ecosystems/f/flow.toml
```diff
@@ -4851,3 +4836,1032 @@ url = "https://github.com/zolrath/flow-remix-next"
...
+url = "https://github.com/Thomatang/pinata-party-frontend"
+
+[[repo]]
+url = "https://github.com/SongNFT/FlowBlockchainNFTMarketplace"
+
+[[repo]]
+url = "https://github.com/thelebster/flow-dapper-wallet-sample-app"
+
+[[repo]]
+url = "https://github.com/tanukixyz/tanuki"
```
jubos, line 5624: This one is general as opposed to Flow, so should probably just go into the General ecosystem.

data/ecosystems/f/flow.toml
```diff
@@ -4851,3 +4836,1032 @@ url = "https://github.com/zolrath/flow-remix-next"
...
+url = "https://github.com/MatrixLabsTech/MatrixUnityTool"
+
+[[repo]]
+url = "https://github.com/interflowrepo/interflow"
+
+[[repo]]
+url = "https://github.com/sibelly/flow-fundamentals"
+
+[[repo]]
+url = "https://github.com/portto/wallet-web-ui"
```
jubos, line 5486: This is a general purpose wallet, so shouldn't be a direct sub repo of Flow.

jubos (changes requested):
I found a few repos that are more general than just directly Flow repos.  Can you do a once over on these and ensure they a directly building on Flow?

data/ecosystems/f/flow.toml
muttoni: Thanks, did a once over

muttoni:
Thanks for the review - did another run through and removed some additional repos. Is this good to merge now? :)

jubos:
Looks good thanks

PR #477, opened 2023-03-21
Added Projects

GiupiDeLuca:
I added the following projects: 
Mimo, UniFi, ZoomSwap, MinMax Finance, Parrot Defi, Magic Land, WOWswap, Cyclone, StarCrazy


data/ecosystems/c/cyclone.toml
```diff
@@ -0,0 +1,46 @@
+# Ecosystem Level Information
+title = "Cyclone"
```
eherrerosj, line 2: should also be added as a `sub_ecosystem` under IoTeX

data/ecosystems/m/magiclandfinance.toml
```diff
@@ -0,0 +1,8 @@
+# Ecosystem Level Information
+title = "Magic Land"
```
eherrerosj, line 2: should also be added as a `sub_ecosystem` under IoTeX

data/ecosystems/p/parrotdefi.toml
```diff
@@ -0,0 +1,8 @@
+# Ecosystem Level Information
+title = "Parrot DeFi"
```
eherrerosj, line 2: should also be added as a `sub_ecosystem` under IoTeX

data/ecosystems/w/wowswap.toml
```diff
@@ -0,0 +1,52 @@
+# Ecosystem Level Information
+title = "WOWswap"
```
eherrerosj, line 2: should also be added as a `sub_ecosystem` under IoTeX

data/ecosystems/z/zoomswap.toml
```diff
@@ -0,0 +1,28 @@
+# Ecosystem Level Information
+title = "ZoomSwap"
```
eherrerosj, line 2: should also be added as a `sub_ecosystem` under IoTeX

eherrerosj (changes requested):
Thanks for the PR @GiupiDeLuca! Left a few comments. Mainly 3 changes are needed:
1. adding all these projects under IoTeX toml so that it is seen as a parent 
2. removing duplicates
3. fixing one of the project titles

GiupiDeLuca:
Thanks for the feedback @eherrerosj - I think I should have everything now. 
- I deleted the duplicates
- Added the projects under iotex
- Fixed the wrong title

GiupiDeLuca:
@eherrerosj I have 2 checks failing, would you be able to help with those? Not sure what I did wrong. 😆

eherrerosj:
> @eherrerosj I have 2 checks failing, would you be able to help with those? Not sure what I did wrong. 😆

Thanks for the update. Not sure by 97fe is failing (can ignore), but f881It is failing because it's now a duplicate (seems like we have it already in our db), you can remove that addition then

eherrerosj:
yo can also bring the changes from your other PR to this one

GiupiDeLuca:
@eherrerosj I removed f881, I see there is minimax-finance, but it seems to be a different project than the MinMax Finance I added. Anyway, I also removed MinMax from the list of iotex sub_ecosystems for now, maybe later on we can check if these two are in fact different project with an almost identical name, or they are indeed the same. Thanks again for the feedback

eherrerosj:
thank you @GiupiDeLuca !

PR #480, opened 2023-04-04
update

letmejustputthishere:


eherrerosj:
Nice large contribution. Could you please add the ecosystems to the `sub_ecosystems` list of the parent ecosystems when it proceeds. Example: "DFINITY Side Projects" should be a sub_ecosystem of DFINITY

eherrerosj (changes requested):
Nice large contribution. Could you please add the ecosystems to the sub_ecosystems list of the parent ecosystems when it proceeds. Example: "DFINITY Side Projects" should be a sub_ecosystem of DFINITY

letmejustputthishere:
> Nice large contribution. Could you please add the ecosystems to the sub_ecosystems list of the parent ecosystems when it proceeds. Example: "DFINITY Side Projects" should be a sub_ecosystem of DFINITY

`DFINITY Side Projects` is a `sub_ecosystem` of `DFINITY`. The name of that org is `DFINITY`
https://github.com/letmejustputthishere/crypto-ecosystems/blob/4cedda0b7d2273f11f28ea99d31c0b6ca7926e3a/data/ecosystems/d/dfinity-side-projects.toml#L1

thus it was added under that `name`/`title` to the `sub_ecosystems` here
https://github.com/letmejustputthishere/crypto-ecosystems/blob/4cedda0b7d2273f11f28ea99d31c0b6ca7926e3a/data/ecosystems/i/internet-computer.toml#L76

Or am I misunderstanding you 🤔

letmejustputthishere:
btw, what is the naming convention here if two ecosystems have the same name? e.g. `dfinity` and `dfinity-side-projects` are both named `DFINITY`. so far we always took the `name` set by the github organisation as the `title` of the corresponding `toml` file

eherrerosj:
The title of [this](https://github.com/letmejustputthishere/crypto-ecosystems/blob/4cedda0b7d2273f11f28ea99d31c0b6ca7926e3a/data/ecosystems/d/dfinity-side-projects.toml#L1) subecosystem should be "DFINITY Side Projects" and not "DFINITY".

Then, "DFINITY Side Projects" should be added as a sub_ecosystem under the DFINITY ecosystem definition [here](https://github.com/letmejustputthishere/crypto-ecosystems/blob/4cedda0b7d2273f11f28ea99d31c0b6ca7926e3a/data/ecosystems/d/dfinity.toml#L2).

I don't recommend always adding the org name as the ecosystem name. I would prioritize clarity and avoid name clashes.

letmejustputthishere:
@eherrerosj is this good to merge now? 😇 

> I would prioritize clarity and avoid name clashes.
we automate this entire process, is there a naming convention you would recommend to avoid duplicates?

eherrerosj (changes requested):
Hi @letmejustputthishere, thank you for your PR. Before we proceed with merging, please consider the following revisions:

When creating ecosystems from repos within organizations, ensure this is done programmatically and selectively. Modify the script to be more restrictive in determining when to generate new ecosystems.

Typically, we establish ecosystems when there is a potential community behind the project, often indicated by a Twitter/Discord account and a dedicated website. For smaller ecosystems with only a few repos, we prefer to include repo URLs within the parent ecosystem.

Example: The "55foundry" org has only one relevant repo for ICP (DIP20). There seems to be no ICP base or community for this project, so a separate ecosystem should not be created.

Additionally, be cautious when adding organization links to the "github_organizations" list in the toml file. Our ETL process will automatically add all repos later on from the specified organization to its corresponding ecosystem ("org expansion"). Due to this, we must be vigilant when incorporating new organizations into ecosystems.

Please let me know if you have any questions.

data/ecosystems/a/actions-marketplace-validations.toml
```diff
@@ -0,0 +1,6 @@
+title = "actions-marketplace-validations"
+sub_ecosystems = []
+github_organizations = ["https://github.com/actions-marketplace-validations"]
```
eherrerosj, line 3: As a critical example of what my main comment was referring to. This org has 18.6k repos, and since you'd be adding it as an org to ICP, all those repos would get linked to ICP, and not only the FleekHQ one, like it should.

Again, better to keep the repos in the ICP toml and crete ecosystems only when it makes sense. And only add orgs when you are sure all its repos should belong to that sub-eco.

eherrerosj (changes requested):

data/ecosystems/a/actions-marketplace-validations.toml
letmejustputthishere: thank you for those insights! that helps a lot :) i think i will manually review the creation of new ecosystems as it seems like its something thats not easy to automate

letmejustputthishere:
why aren't you using the orgs github account name as a unique identifier when referencing sub ecosystems?

letmejustputthishere:
> Additionally, be cautious when adding organization links to the "github_organizations" list in the toml file. Our ETL process will automatically add all repos later on from the specified organization to its corresponding ecosystem ("org expansion"). Due to this, we must be vigilant when incorporating new organizations into ecosystems.

this part i don't understand, can you elaborate a bit? the definition of `github_organization` is still a bit unclear to me, one of your colleagues said 

> A github org = a github organization that belongs to Dfinity and is not its own crypto project (example: https://github.com/dfinity)

so far i was under the impression that this means the org has to be part of dfinity somehow, but looking at other ecosystems that doesn't seem to be the case.

when do i add an org to the ICs `github_organizations`? when adding it, can it have it's own ecosystem file? what is the definition of a "crypto project" in the above quote?

data/ecosystems/a/actions-marketplace-validations.toml
eherrerosj: totally. In fact, we have been experimenting with ML to classify repos and ecosystems and it's not an easy task. We are in the process of semi-automating a part of it (eg cases where the classification is obvious)

eherrerosj:
> why aren't you using the orgs github account name as a unique identifier when referencing sub ecosystems?

Because an ecosystem might have more than one org attached to it. Also, an org can belong to multiple ecosystems

eherrerosj:
> > Additionally, be cautious when adding organization links to the "github_organizations" list in the toml file. Our ETL process will automatically add all repos later on from the specified organization to its corresponding ecosystem ("org expansion"). Due to this, we must be vigilant when incorporating new organizations into ecosystems.
> 
> this part i don't understand, can you elaborate a bit? the definition of `github_organization` is still a bit unclear to me, one of your colleagues said
> 
> > A github org = a github organization that belongs to Dfinity and is not its own crypto project (example: https://github.com/dfinity)
> 
> so far i was under the impression that this means the org has to be part of dfinity somehow, but looking at other ecosystems that doesn't seem to be the case.
> 
> when do i add an org to the ICs `github_organizations`?
when you see that all the repos created under that org should belong to the org's parent ecosystem

> when adding it, can it have it's own ecosystem file? what is the definition of a "crypto project" in the above quote?
when adding it, it can and should be added to its own ecosystem if it feels like it currently has or could have its own community in the future. A good rule of thumb is looking for a website and socials (twitter, discord, similars)

letmejustputthishere:
lets take [this](https://github.com/hashquark-io?) org for example.
i'm going to create a new ecosystem file `hasquark-io.toml` for it, as it has its own website and twitter.

i will leave the `github_organizations` list in `hashquark-io.toml` empty, as not all repositories of that organisation are related to the internet computer. i will add all IC related repositories to `hashquark-io.toml`. i will reference the toml in the `sub_ecosystems` list of `internet-computer.toml`. 

is that correct?

```
// hashquark-io.toml
title = "HashQuark"
sub_ecosystems = []
github_organizations = ["https://github.com/hashquark-io"]
[[repo]]
url = "https://github.com/hashquark-io/blockchart"

[[repo]]
url = "https://github.com/hashquark-io/nnsExplorer"
```

PR #485, opened 2023-04-21
Collectif DAO, liquid staking protocol on Filecoin added

kmadorin:


data/ecosystems/f/filecoin.toml
```diff
@@ -3,6 +3,7 @@ title = "Filecoin"
 
 sub_ecosystems = [
   "Application Research Group",
+  "Collectif DAO"
```
eherrerosj, line 6: a "," is missing at the end of this line, can you please fix it?

eherrerosj (changes requested):
thanks for the contribution, can you please fix the syntax error that I pointed out?

PR #490, opened 2023-04-27
Update icp 2023 04 27

letmejustputthishere:


data/ecosystems/d/dantenetwork.toml
```diff
@@ -0,0 +1,6 @@
+title = "dantenetwork"
+sub_ecosystems = []
+github_organizations = ["https://github.com/dantenetwork"]
```
eherrerosj, line 3: we have a process that finds all repos under orgs and maps those repos to the parent ecosystem. I see that this org has 60+ repos but only `canister_contracts` should belong to IC. Could you add only that repo to icp? (Instead of this whole ecosystem)

data/ecosystems/d/defi-tools.toml
```diff
@@ -0,0 +1,6 @@
+title = "DeFi-Tools"
+sub_ecosystems = []
+github_organizations = ["https://github.com/DeFi-Tools"]
```
eherrerosj, line 3: we have a process that finds all repos under orgs and maps those repos to the parent ecosystem. I see that this org has 120+ repos but only `sudograph` should belong to IC. Could you add only that repo to icp? (Instead of this whole ecosystem)

data/ecosystems/d/departurelabsic.toml
```diff
@@ -0,0 +1,51 @@
+title = "Departure Labs"
```
eherrerosj, line 1: duplicate of https://github.com/electric-capital/crypto-ecosystems/blob/master/data/ecosystems/d/departure-labs.toml
can remove this file

data/ecosystems/d/dfinity-education.toml
```diff
@@ -0,0 +1,24 @@
+title = "Internet Computer Education"
```
eherrerosj, line 1: All these repos might already be under ICP main ecosystem file. Example: https://github.com/electric-capital/crypto-ecosystems/blob/bbdf537935ae1d48227c96a995359f581ab51fd2/data/ecosystems/i/internet-computer.toml#L2773

data/ecosystems/d/dfinity-side-projects.toml
```diff
@@ -0,0 +1,93 @@
+title = "DFINITY Side Projects"
```
eherrerosj, line 1: All these repos might already be under ICP main ecosystem file

data/ecosystems/d/digifun-dao.toml
```diff
@@ -0,0 +1,9 @@
+title = "DigiFun-DAO"
+sub_ecosystems = []
+github_organizations = ["https://github.com/DigiFun-DAO"]
+[[repo]]
+url = "https://github.com/DigiFun-DAO/frontend-mobile"
+
+[[repo]]
+url = "https://github.com/DigiFun-DAO/frontend"
```
eherrerosj, line 8: All these repos might already be under ICP main ecosystem file

data/ecosystems/d/dinternet-computer.toml
```diff
@@ -0,0 +1,6 @@
+title = "dinternet-computer"
+sub_ecosystems = []
+github_organizations = ["https://github.com/dinternet-computer"]
+[[repo]]
+url = "https://github.com/dinternet-computer/rust_hello"
```
eherrerosj, line 5: All these repos might already be under ICP main ecosystem file

data/ecosystems/d/distive.toml
```diff
@@ -0,0 +1,9 @@
+title = "Distive"
+sub_ecosystems = []
+github_organizations = ["https://github.com/Distive"]
+[[repo]]
+url = "https://github.com/Distive/Distive"
+
+[[repo]]
+url = "https://github.com/Distive/distive-dashboard"
```
eherrerosj, line 8: All these repos might already be under ICP main ecosystem file

eherrerosj (changes requested):
hey thanks for the contribution. I'm a bit confused with some of the changes you are proposing. Main thoughts:
1. You created a lot of new ecosystems with orgs that have at least a repo that should be part of IC. This is correct. However, in many cases we already had that repo as part of IC but after isolating into its own eco, you didn't remove those duplicates from the IC ecosystem file
2. As a continuation of 1), you added each ecosystem's org when possible and the repo that belongs to IC. The idea is great, but there's a caveat. We have a process that automatically finds all repos under the given org and attaches them to all the parent ecosystems that the ecosystem's org has. In order to add only some repos of an org, we simply add those repos to the intended parent ecosystem. Please do the same for orgs that don't only work on top of IC

letmejustputthishere:
> You created a lot of new ecosystems with orgs that have at least a repo that should be part of IC. This is correct. However, in many cases we already had that repo as part of IC but after isolating into its own eco, you didn't remove those duplicates from the IC ecosystem file

thank you for highlighting this, i wasn't aware that this is the practice. i will remove duplicates from the `ic` ecosystem if a new ecosystem is being created for an org

eherrerosj:
> > You created a lot of new ecosystems with orgs that have at least a repo that should be part of IC. This is correct. However, in many cases we already had that repo as part of IC but after isolating into its own eco, you didn't remove those duplicates from the IC ecosystem file
> 
> thank you for highlighting this, i wasn't aware that this is the practice. i will remove duplicates from the `ic` ecosystem if a new ecosystem is being created for an org

Sounds good. Another thing I forgot to mention is that, as a rule of thumb, we tend to isolate as ecosystems those orgs that have a website and/or twitter / discord in their github profile.

PR #492, opened 2023-04-30
add new repos, remove and rename old repos

jlopp:
adding more lightning repositories
removing deleted bitcoin repositories
renaming moved bitcoin repositories


data/ecosystems/b/bitcoin.toml
```diff
@@ -1811,18 +1817,6 @@ url = "https://github.com/ElementsProject/reserves"
 [[repo]]
 url = "https://github.com/ElementsProject/rust-elements"
 
-[[repo]]
-url = "https://github.com/ElementsProject/rust-secp256k1-zkp"
```
eherrerosj, line 1815: even tho these are forks of `BlockstreamResearch` repos, you can keep them because we have an algorithm for fork detection and tagging to deduplicate commit counts

eherrerosj (changes requested):
thanks for the contribution, I left a couple comments, can you please work on the changes and we will have these merge? If we merge within the next couple days, the data will be reflected in May's update of developerreport.com

data/ecosystems/b/bitcoin.toml
```diff
@@ -1222,9 +1234,6 @@ url = "https://github.com/decentralized-identity/ion"
 [[repo]]
 url = "https://github.com/decentralized-identity/uni-registrar-driver-did-btcr"
 
-[[repo]]
-url = "https://github.com/DeFiCh/app"
```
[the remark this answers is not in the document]
jlopp: Because it no longer exists; it's an old URL for what is now https://github.com/BirthdayResearch/defichain-app

DefiChain is a subecosystem in bitcoin.toml

Looks like that subecosystem has https://github.com/BirthdayResearch/app which is no longer valid either; I'll update it to https://github.com/BirthdayResearch/defichain-app

data/ecosystems/b/bitcoin.toml
jlopp: These aren't forks of BlockstreamResearch repos, they were moved FROM the ElementsProject repository TO the BlockstreamResearch repository.

eherrerosj (approved):
thanks for contributing @jlopp ! all looks good, merging

PR #493, opened 2023-05-03
Add Eluvio as Polkadot sub-ecosystem

elv-nate:


data/ecosystems/p/polkadot.toml
```diff
@@ -83,6 +83,7 @@ sub_ecosystems = [
   "DotOracle Network",
   "dTrade",
   "Edgeware",
+	"Eluvio",
```
eherrerosj, line 86: Hey @elv-nate, can you keep the indented formatting like the rest of the lines? You can use 2 spaces I believe

eherrerosj (changes requested):

data/ecosystems/p/polkadot.toml
elv-nate: Ah whoops, yeah just did. I have my editor settings displaying a tab as 2 spaces, so I didn't notice the difference.

eherrerosj (approved):
Thanks!

PR #496, opened 2023-05-03
Tidy up for flow orgs

franklywatson:
Cleanup to remove orgs that have since been removed, or repos that are deleted. Also moved a bunch of individual repo listings into orgs

### DIFF
--- a/data/ecosystems/f/flow.toml
+++ b/data/ecosystems/f/flow.toml
@@ -5,7 +5,6 @@ sub_ecosystems = [
   "33 Labs",
   "Arlequin: The Artists Metaverse",
   "Basicbeasts",
-  "BSideGames (Chain Monsters)",
   "Dynamiclabs",
   "ELASTECH Services",
   "Emerald City DAO",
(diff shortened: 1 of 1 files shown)


data/ecosystems/f/flow.toml
```diff
@@ -69,27 +82,6 @@ github_organizations = [
 ]
 
 # Repositories
-[[repo]]
-url = "https://github.com/0xAlchemist/Athletaverse"
```
eherrerosj, line 73: why are you requesting to remove these ones?

franklywatson: You have linked one item, but spoke in plural. Was there more than one? 

For the one shown looks like i removed that in error. Let me fix that

eherrerosj: Yeah, I'm referring to all the repos that you removed.

In your PR, there's 400 flow repos that are gone when I compare them with the original file (~1900 vs ~1500). Am I missing something?

franklywatson: Yeah I moved a bunch of the individual items in the lower list into orgs where I know these are Flow repos. However, from looking at the PR there's something bad about the way I edited this which makes the change diff look really horrible and makes it hard to see what I am explaining. I will cancel this and make a cleaner PR so you can what was changed more directly

PR #503, opened 2023-05-12
Add FENIX ecosystem

joeblau:
Just watched @mariashen give a talk at BuildETH. Following up 🫡

### DIFF
--- a/data/ecosystems/e/ethereum.toml
+++ b/data/ecosystems/e/ethereum.toml
@@ -865,7 +865,6 @@ sub_ecosystems = [
 github_organizations = [
   "https://github.com/0xPikapool",
   "https://github.com/Aloha-Open-Wifi",
-  "https://github.com/atomizexyz",
   "https://github.com/chelofinance",
   "https://github.com/commons-stack",
   "https://github.com/crytic",
@@ -6795,9 +6794,6 @@ url = "https://github.com/LaTribuWeb3/dex-history-fetcher"
 
 [[repo]]
 url = "https://github.com/LaTribuWeb3/fei-protocol-core"
-
-[[repo]]
-url = "https://github.com/atomizexyz/fenix"
 
 [[repo]]
 url = "https://github.com/LaTribuWeb3/monitor-backend"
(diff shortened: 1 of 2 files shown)


data/ecosystems/f/fenix.toml
```diff
@@ -0,0 +1,22 @@
+# Ecosystem Level Information
+title = "Fenix Protocol"
```
eherrerosj, line 2: hey @joeblau, thanks for the contribution. I see the Fenix is built on Ethereum. If so, please add "Fenix Protocol" to the list of sub_ecosystems on file `ethereum.toml` under `data/ecosystems/e`

eherrerosj (changes requested):
hey @joeblau, thanks for the contribution. I see the Fenix is built on Ethereum. If so, please add "Fenix Protocol" to the list of sub_ecosystems on file ethereum.toml under data/ecosystems/e

joeblau:
FENIX will exist on 13 chains... Ethereum, BSC, Polygon, Avalanche, EthereumPOW, Moonbeam, Evmos, Fantom, DogeChain, OKChain, Arbitrum, Optimism, and PulseChain. Our community views the EVM as the value proposition, Ethereum is just one of the chains we're launching on.

That being said I've updated the PR.

data/ecosystems/e/ethereum.toml
```diff
@@ -6794,6 +6796,9 @@ url = "https://github.com/LaTribuWeb3/dex-history-fetcher"
 [[repo]]
 url = "https://github.com/LaTribuWeb3/fei-protocol-core"
 
+[[repo]]
+url = "https://github.com/atomizexyz/fenix"
```
eherrerosj, line 6800: you can remove this repo, since it's already in the Ethereum's child "Fenix" ecosystem

eherrerosj (changes requested):
We only need the ecosystem name in Ethereum, our scripts do the rest to come up with the entire tree of relations.

Wrt to the multi-chain compatibility. We have this internal concept of "primary" and "supported" ecosystems. We consider a relation between ecos to be primary when it has a currency / gov token in that network or it was originally built there and deployed afterwards to other chains. If Ethereum happen to not be the case of primary to Fenix please lmk

eherrerosj (approved):
Perfect, merging! Thank you for the contribution

PR #504, opened 2023-05-12
Update xdc.toml

ChainGamer:
XDC Network consists of many different entities building across the world.  These updates reflect the repositories that need to be scanned in order to retrieve accurate data of developer activity.


ChainGamer:
I didn't realize we just needed to add the sub orgs and the repos would automatically be added.

ChainGamer (commented):
Updated

ChainGamer (commented):
Updated

PR #516, opened 2023-06-05
update icp ecosystem

letmejustputthishere:


eherrerosj (approved):
Great work! Thanks a lot for all the iterations you did. Feels like we got to a great spot. Ideally we can create a monthly PR like this one and we ensure ICP/DFINITY is always up to date.

Will merge this now 🙏🏻

PR #527, opened 2023-06-13
added Arda and update for May

schimih:
Added:

- Arda;
- new repos from ecosystem;


eherrerosj (approved):
merging this, thanks for the contribution

PR #532, opened 2023-06-24
Update rootstock.toml

ezequiel-rodriguez:


eherrerosj (approved):
thanks for the contribution!

PR #545, opened 2023-07-11
Add Struct Finance to Avalanche Sub Eco

eXale1n:
Adds Struct Finance

- Created  `struct-finance.toml` in data/ecosystems/s
- Added `Struct Finance` to `avalanche.toml`


eherrerosj (approved):
nice addition, I've just removed the isolated repo in Avalanche because it was not necessary

PR #548, opened 2023-07-11
Add DeFi Kingdoms repos

magnusironroot:
Adds DeFi Kingdoms repos


eherrerosj:
Defi Kingdoms is already part of Avalanche. You can see it [here](https://github.com/electric-capital/crypto-ecosystems/blob/master/data/ecosystems/a/avalanche.toml#L44)

PR #561, opened 2023-07-30
Add Kodadot to various L2 subecosystems

yangwao:
We are deploying NFT Explorer on other L2s in Ethereum too

More to verify are here on alpha deployment - https://canary.kodadot.xyz/

### DIFF
--- a/data/ecosystems/a/aleph-zero.toml
+++ b/data/ecosystems/a/aleph-zero.toml
@@ -1,7 +1,7 @@
 # Ecosystem Level Information
 title = "Aleph Zero"
 
-sub_ecosystems = ["Artzero"]
+sub_ecosystems = ["Artzero", "Kodadot"]
 
 github_organizations = [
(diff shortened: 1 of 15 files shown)


eherrerosj:
A couple fixes needed here @yangwao:
1. Kodadot should be a sub-ecosystem of the supported chains, not the other way around
2. The name used under the `sub_ecosystems` list has to be exactly the same as the ecosystem names in their corresponding toml files, and it is case-sensitive, that's why this is failing

yangwao:
> 1. Kodadot should be a sub-ecosystem of the supported chains, not the other way around

what's the format for that? I should add KodaDot to sub-eco of other chains if I get it right?

> 2\. and it is case-sensitive, that's why this is failing

Ah, I thought after wards I've submitted it, can proceed with fix

yangwao:
Hey @eherrerosj I see green checks now! :)

eherrerosj (approved):
thanks for the contribution! merging now

PR #562, opened 2023-07-30
Update cardano's data with some well-known projects / organizations.

KtorZ:
Cheers ❤️

### DIFF
--- a/data/ecosystems/c/cardano.toml
+++ b/data/ecosystems/c/cardano.toml
@@ -32,11 +32,29 @@ sub_ecosystems = [
 
 github_organizations = [
   "https://github.com/aadafinance",
+  "https://github.com/aiken-lang",
   "https://github.com/armada-alliance",
+  "https://github.com/berry-pool",
+  "https://github.com/blinklabs-io",
+  "https://github.com/bloxbean",
+  "https://github.com/cardano-community",
   "https://github.com/cardano-foundation",
   "https://github.com/CardanoSolutions",
+  "https://github.com/fivebinaries",
+  "https://github.com/geniusyield",
+  "https://github.com/GimbaLabs",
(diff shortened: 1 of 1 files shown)


data/ecosystems/c/cardano.toml
```diff
@@ -407,12 +512,6 @@ url = "https://github.com/CardanoUmbrella/awesome-cardano"
 [[repo]]
 url = "https://github.com/cffls/pycardano"
 
-[[repo]]
-url = "https://github.com/cloudstruct/go-cardano-submit-api"
-
-[[repo]]
-url = "https://github.com/cloudstruct/go-ouroboros-network"
```
eherrerosj, line 414: why do you propose to remove these 2 repos?

eherrerosj (changes requested):
Overall looks good! I would like to know why you are proposing to remove https://github.com/cloudstruct/go-cardano-submit-api and https://github.com/cloudstruct/go-ouroboros-network from Cardano taxonomy

data/ecosystems/c/cardano.toml
KtorZ: CloudStruct has become BlinkLabs, so I added the BlinkLabs equivalent instead 👍

KtorZ: More specifically: 

| Old | New | 
| --- | --- | 
| https://github.com/cloudstruct/go-ouroboros-network | https://github.com/blinklabs-io/gouroboros |
| https://github.com/cloudstruct/go-cardano-submit-api | https://github.com/blinklabs-io/tx-submit-api |

You'll see that Github even redirects you properly if you click the old links.

eherrerosj (approved):
LGTM. Thanks for the contribution. Merging now

PR #567, opened 2023-08-08
Update iotex.toml

GiupiDeLuca:


data/ecosystems/i/iotex.toml
```diff
@@ -1085,3 +1124,126 @@ url = "https://github.com/machinefi/w3bstream_compilets_demo"
...
+url = "https://github.com/mimoprotocol/mimo-analytics"
+
+[[repo]]
+url = "https://github.com/mimoprotocol/mimox-contracts"
+
+[[repo]]
+url = "https://github.com/mimoprotocol/mimo-automator"
+
+[[repo]]
+url = "https://github.com/mimoprotocol/mimo-contract"
```
eherrerosj, line 1249: you can remove all the mimoprotocol repos because `Mimo Defi` is a sub ecosystem already part of IoTeX and therefor we are already accounting for the activity of all its repos and attributing it to IoTeX

data/ecosystems/i/iotex.toml
```diff
@@ -1085,3 +1124,126 @@ url = "https://github.com/machinefi/w3bstream_compilets_demo"
...
+url = "https://github.com/machinefi/oura-demo-v2"
+
+[[repo]]
+url = "https://github.com/machinefi/w3bstream-vscode-extension-assemblyscript"
+
+[[repo]]
+url = "https://github.com/machinefi/w3bstream-vscode-extension-go"
+
+[[repo]]
+url = "https://github.com/machinefi/w3bstream-vscode-extension-rust"
```
eherrerosj, line 1180: might be worth creating a new eco for "MachineFi Lab" and have its org link added to its github_organization field. And then add that new eco as a sub eco to IoTeX. That way we always take all existing and new repos of that org and attribute to IoTeX, wdyt?

eherrerosj (changes requested):
left a couple comments

eherrerosj (approved):

PR #571, opened 2023-08-23
Add three new Stacks repos

kenrogers:


eherrerosj (approved):
thanks for the contribution, merging

PR #591, opened 2023-09-12
Adding EOS organizations and repositories

nsjames:
Hello!

I'd like to add some more recent data to the EOS ecosystem please. 

There are two primary orgs that are missing:
- [EOS Network Foundation](https://github.com/orgs/eosnetworkfoundation/) - EOS-specific code and EOS EVM
- [AntelopeIO](https://github.com/AntelopeIO) - The Antelope protocol (the successor to EOSIO, and the tech that the foundation develops to power EOS)

I've also added the most important repositories from both of them as well.


eherrerosj (approved):
TY for the contribution, merging

PR #612, opened 2023-10-05
Update Solady repo

atarapra:


eherrerosj:
Closing since `Vectorized` is a user and not a current / potential ecosystems

data/ecosystems/s/solady.toml
```diff
@@ -0,0 +1,10 @@
+# Ecosystem Level Information
+title = "Solady"
```
eherrerosj, line 2: Could you also add Solady as a sub_ecosystem of Ethereum and remove the repo link from Ethereum?

eherrerosj (changes requested):
Thanks for the contribution and apologies for the oversight closing the PR before.

Just a couple of small changes and we are good to go.

atarapra:
@eherrerosj you can review now. updated all changes.

PR #617, opened 2023-10-14
Freshen up Polkadot repos

themacexpert:
This updates Polkadot repos. It does not touch Polkadot subecosystems.


eherrerosj (changes requested):
hey @themacexpert, you think we could add a missing tag instead of removing repos that don't exist anymore publicly? We prefer to keep non-existant repos so we know they existed at some point and we can keep their commits. 

Example: https://github.com/electric-capital/crypto-ecosystems/blob/master/data/ecosystems/h/helios-labs.toml#L11

themacexpert:
> hey @themacexpert, you think we could add a missing tag instead of removing repos that don't exist anymore publicly? We prefer to keep non-existant repos so we know they existed at some point and we can keep their commits.
> 
> Example: https://github.com/electric-capital/crypto-ecosystems/blob/master/data/ecosystems/h/helios-labs.toml#L11

Done!

eherrerosj (approved):
thank you!

PR #631, opened 2023-10-22
added RadixCharts as radix-dlt sub_ecosystem

NellySayon:


data/ecosystems/r/radix-dlt.toml
```diff
@@ -1,7 +1,7 @@
 # Ecosystem Level Information
 title = "Radix DLT"
 
-sub_ecosystems = []
+sub_ecosystems = [RadixCharts]
```
eherrerosj, line 4: please add the double quotes around the subecosystem name, thx!

eherrerosj (changes requested):

data/ecosystems/r/radixcharts.toml
```diff
@@ -0,0 +1,10 @@
+# Ecosystem Level Information
+title = "RadixCharts"
+
+sub_ecosystems = []
+
+github_organizations = ["https://github.com/NellySayon"]
```
eherrerosj, line 6: this link corresponds to a github user account, not an org, so we can't use it, please remove

data/ecosystems/r/radixcharts.toml
```diff
@@ -0,0 +1,10 @@
...
+
+# Repositories
+[[repo]]
+url = "https://github.com/NellySayon/ecobadge"
```
eherrerosj, line 10: why is the repo and project name so different from the subecosystem name you are adding? (ecobadge vs RadixCharts)

eherrerosj (changes requested):

eherrerosj:
Hi @NellySayon, we wanted to give you a heads-up that we are planning to freeze the data set that will be used for the analysis during the next few days. If you want your changes to be included and have an impact on this year’s report, please ensure that the requested modifications are made and merged prior to the data freeze.

Let's aim to have your PR finalized as soon as possible to make it into this window. If you have any questions or need further clarification, feel free to reply here or reach out directly.

data/ecosystems/r/radix-dlt.toml
eherrerosj: fixed it

data/ecosystems/r/radixcharts.toml
eherrerosj: fixed it

PR #636, opened 2023-10-24
Update sui.toml

georgescharlesbrain:
added some tags

I want to use the sui.toml file as a lookup tool for quality Sui Move repos. What I am still missing now are:
- tags for categorization, language used (move, typescript, ...)
- a curation of quality repos vs test repos


eherrerosj:
like mentioned in the other PR, let's try to think how to scale up a tag system that works for all repos. Otherwise it will be inconsistent

PR #637, opened 2023-10-24
Delete data/ecosystems/s/sui-name-service-sns.toml

georgescharlesbrain:
This should just be part of the sui ecosystem.  
Otherwise we can make a toml for each Dapp...


eherrerosj:
@georgescharlesbrain, you are right, Sui Name Service (SNS) should be a child of Sui. I've added it internally right now after your suggestion. I'll close this PR now and my change will show up in the next data update

PR #639, opened 2023-10-24
Update sui.toml

georgescharlesbrain:
tag forks


eherrerosj:
We have an internal fork detection algorithm. Like my previous comment, we prefer to add structural changes to repos at scale

PR #654, opened 2023-11-10
Create drop-wireless.toml

GiupiDeLuca:


eherrerosj (approved):
thx!

data/ecosystems/d/drop-wireless.toml
```diff
@@ -0,0 +1,25 @@
+# Ecosystem Level Information
+title = "drop-wireless"
```
eherrerosj, line 2: @GiupiDeLuca, can we use "Drop Wireless" for the tittle instead?

eherrerosj (changes requested):
small req

data/ecosystems/d/drop-wireless.toml
GiupiDeLuca: Sure, will fix now!

PR #657, opened 2023-11-15
Add Axelar Ecosystems

benjamin852:
Added in `sub_ecosystems` for Axelar network


eherrerosj (changes requested):
Could you remove the empty ecos? You can add them back later when/if you find its orgs / repos. After this it should be good to merge

eherrerosj (approved):
Nice contribution! Thank you. Merging now

PR #660, opened 2023-11-20
Add InterPlanetary Consensus

jsoares:
Adds a number of missing repos, mainly related to Filecoin's [InterPlanetary Consensus](https://filecoin.io/blog/posts/achieving-breakthrough-scalability-with-interplanetary-consensus-ipc/).

It's not entirely clear whether it makes sense to duplicate repos in the parent and sub ecosystems, but that seems to be the case already for other entries.

### DIFF
--- a/data/ecosystems/f/filecoin.toml
+++ b/data/ecosystems/f/filecoin.toml
@@ -8,7 +8,7 @@ sub_ecosystems = [
   "InterPlanetary Consensus",  
   "NFT Storage",
   "Slate Engineering Services",
-  "Web3 Storage"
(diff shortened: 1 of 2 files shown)


eherrerosj (changes requested):
Could you remove the repos that you have in the new toml file from filecoin? I know it happens in other tomls but better to keep it as clean as possible. The devs from the new toml will get rolled up to filecoin anyways :)

jsoares:
@eherrerosj deduplicated as requested

eherrerosj (approved):
thx for the contribution! merging

PR #663, opened 2023-11-21
Added missing org for IncrementFi

franklywatson:


data/ecosystems/f/flow.toml
```diff
@@ -75,6 +75,7 @@ github_organizations = [
   "https://github.com/FlowFans-Learning",
   "https://github.com/letthemusicpay",
   "https://github.com/LiquidityTokenLabs",
+  "https://github.com/orgs/IncrementFi",
```
eherrerosj, line 78: can you replace this org URL with adding IncrementFi as a subecosystem of Flow (see list at the beginning of this same file)?

eherrerosj (changes requested):
Added a comment

data/ecosystems/f/flow.toml
btspoony, line 1: I found that IncrementFi already exists in the sub-ecosystems, but just missing some repos:
https://github.com/electric-capital/crypto-ecosystems/blob/master/data/ecosystems/i/increment-finance.toml

eherrerosj: @franklywatson could you please adjust the PR to add the missing repos to the existing toml?

eherrerosj (changes requested):

eherrerosj (approved):

PR #678, opened 2023-11-28
add more avalanche ecosystem repos and github orgs

ashucoder9:


ashucoder9:
Fixed with https://github.com/electric-capital/crypto-ecosystems/pull/681

PR #683, opened 2023-12-03
Fix outdated DipDup links

droserasprout:
We have renamed GitHub organization to match dipdup.io domain some time ago. Other links in `dipdup.toml` and `baking-bad.toml` may be outdated too. Are they verified/refreshed on schedule somewhere?

### DIFF
--- a/data/ecosystems/d/dipdup.toml
+++ b/data/ecosystems/d/dipdup.toml
@@ -3,7 +3,7 @@ title = "DipDup"
 
 sub_ecosystems = []
 
-github_organizations = ["https://github.com/dipdup-net"]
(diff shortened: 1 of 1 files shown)


eherrerosj (approved):
thanks!

eherrerosj:
> We have renamed GitHub organization to match dipdup.io domain some time ago. Other links in `dipdup.toml` and `baking-bad.toml` may be outdated too. Are they verified/refreshed on schedule somewhere?

would only get fixed automatically if there's redirections in github (renaming of repos/orgs normally works fine). Otherwise the github refreshers won't know which one is the new name. In those cases we might need to fix those

PR #688, opened 2023-12-07
Added Tuum Tech

kpachhai:


data/ecosystems/h/hedera.toml
```diff
@@ -18,6 +18,7 @@ sub_ecosystems = [
   "Mintmaster Labs",
   "PieFi Platform",
   "SaucerSwap Labs",
+  "Tuum Tech",
```
eherrerosj, line 21: I don't think this ecosystem currently exist in our taxonomy

eherrerosj (changes requested):
I don't see any toml file (ecosystem) with name "Tuum Tech". This means that adding it here would not have any effect. There's 2 options:
1. Remove this line and keep the 2 repos you added below
2. Create a toml file for Tuum Tech and move the repos there. Then you can keep this line

I would suggest the 2nd option given that Tuum Tech does look like an ecosystem in which other devs could build

eherrerosj (approved):
LGTM now, thanks!

PR #696, opened 2023-12-12
Update aptos.toml

zacharyr0th:
Updated to included the open source repos dependent on Aptos-Core that have at least 1 GitHub star.


eherrerosj (changes requested):
Left a comment here and shared a list of a few hundred repos you are adding for which I can't find traces of Aptos usage and will need you to double check

data/ecosystems/a/aptos.toml
```diff
@@ -2,6 +2,15 @@
 title = "Aptos"
 
 sub_ecosystems = [
+<<<<<<< HEAD
```
eherrerosj, line 5: you might wanna solve these conflicts first :)

eherrerosj (changes requested):
Some conflicts need to be resolved first, we are almost there!

eherrerosj (approved):
LGTM now, thanks for the contribution! Merging

PR #701, opened 2023-12-14
Add Kabila organization to hedera.toml

AdrianKBL:


eherrerosj:
thanks for the PR, I'm just going to close it because it will soon conflict with the same addition we did internally and that is not yet reflected in this open source taxonomy. Almost your exact change will be added tomorrow when we do the repo refreshing

PR #710, opened 2023-12-19
Add developer tooling repo for celo

aaronmgdr:
add developer-tooling repo which contains packages split out from the core repo

remove celocli repo (not used)


eherrerosj (approved):
merging now, thanks!

PR #722, opened 2023-12-19
new NEAR repos added (includes forks)

0bserver07:


eherrerosj:
As discussed in telegram, we manually added the corresponding orgs and repos in your submission that are non-archived/non-deleted non-fork repos that are not currently not present in the NEAR taxonomy. Thanks!

PR #738, opened 2023-12-21
Add xNFT Nimbus

thanhlmm:


eherrerosj:
closing since we already have an ecosystem for Nimbus with all its repos (https://github.com/electric-capital/crypto-ecosystems/blob/master/data/ecosystems/n/nimbus-getnimbus.toml)

PR #741, opened 2023-12-21
Adding Just Ape subecosystem & org

Just-Ape-Studios-Admin:
Adding Just Ape name & org

### DIFF
--- a/data/ecosystems/s/solana.toml
+++ b/data/ecosystems/s/solana.toml
@@ -121,6 +121,7 @@ sub_ecosystems = [
   "Jet Protocol",
   "Jito Network",
   "Jupiter Exchange",
+  "Just Ape","
   "Kurobi",
   "Kyber Network",
   "kycDAO",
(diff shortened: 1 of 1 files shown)


data/ecosystems/s/solana.toml
```diff
@@ -121,6 +121,7 @@ sub_ecosystems = [
   "Jet Protocol",
   "Jito Network",
   "Jupiter Exchange",
+  "Just Ape","
```
eherrerosj, line 124: typo

eherrerosj: also, if you add an ecosystem here, there needs to exist a toml file with the same "title". I don't think we have a toml file for Just Ape. You have 2 options:
1. Delete this change and keep only the org change below
2. Create a toml file like just-ape.toml with the org in it and remove the org link from below

eherrerosj (changes requested):
Left a comment

jubos:
Going to close this out.  Please add the JustApe ecosystem repos and reopen when ready.

PR #750, opened 2023-12-21
Add avalanche repos from Constellation hackathon

ashucoder9:
I added a few more repositories that we received from Constellation hackathon submissions. I have appended them at the end as I saw you sorted the list recently and added more repos.

cc @eherrerosj


eherrerosj (approved):
LGTM, thanks!

PR #755, opened 2023-12-21
add solana-mobileconnect

maxims94:


eherrerosj (changes requested):
The URL you provided is a nice addition but it is an Organization URL, not a Repository URL. Could you move it to the organizations list around the top of the file with the other org links plz?

maxims94:
I added both the organization and the repositories.

eherrerosj (approved):
LGTM now, thanks!

PR #758, opened 2023-12-21
Add 5afe as a Safe Ecosystem GitHub organization

germartinez:
Adding https://github.com/5afe as a new GitHub organization on the Safe Ecosystem.


eherrerosj (approved):
nice thx!

PR #760, opened 2023-12-21
Remove duplicate Stockpile repos

joeymeere:
We changed our GitHub org username from "StockpileProtocol" to "StockpileLabs", which created duplicates of the same repos in the list


eherrerosj (approved):
good catch, thanks!

PR #761, opened 2023-12-21
add aori.io repos

anticode-dev:
hi, adding aori.io repositories


eherrerosj (changes requested):
looking good, couple suggestions:
1. Rename title to "Aori"
2. Add "Aori" in the sub-ecosystems list in Ethereum, Arbitrum, Avalanche, Polygon, Binance Smart Chain, and Optimism. Since it is compatible with all those

eherrerosj (approved):
great addition! thanks :)

PR #782, opened 2023-12-21
fix ideal labs repo url

driemworks:
The existing repo url does not exist. This PR adds the correct repo link.


eherrerosj (changes requested):
please keep both original urls and add the new url too. We keep a history of all repos even thought they don't exist anymore. This is because we persist all commits and have an impact in timeseries metrics

driemworks (commented):
I added the old URLs back to the toml

eherrerosj (approved):
thanks!

PR #801, opened 2023-12-22
add solcial to solana

augustin-cheron:
Solcial is a permissionless social network, that lets you invest in people.


eherrerosj (approved):
thx!

PR #811, opened 2023-12-22
Create Vanity-ETH-Generator-Poly

samaeleth:
Vanity-ETH-Generator-Poly

### DIFF
--- a/data/ecosystems/v/Vanity-ETH-Generator-Poly
+++ b/data/ecosystems/v/Vanity-ETH-Generator-Poly
@@ -1,0 +1,25 @@
+# Ecosystem Level Information
+title = "Vanity ETH Generator Poly"
+
+sub_ecosystems = []
+
+github_organizations = ["https://github.com/samaeleth"]
+
+# Repositories
+[[repo]]
+url = "https://github.com/samaeleth/vanity-eth-generator/blob/master/Poly-Vanity-ETH"
+
+[[repo]]
+url = "https://github.com/samaeleth/vanity-eth-generator/blob/master/vue.config.js"
+
+[[repo]]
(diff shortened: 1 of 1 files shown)


eherrerosj (changes requested):
Hey, a few comments:

the link under github organization is a User URL not an Org URL
some of the Repo URLs are not properly formatted. The expected format is like https://github.com/org/repo
Is "Vanity ETH Generator Poly" an organization / ecosystem? If so, with what chains is it compatible? Please add "Vanity ETH Generator Poly" as sub-ecosystems of those compatible ecosystems

samaeleth:
followed all your recommendations. thank you.

eherrerosj:
Please add "Vanity ETH Generator Poly" as sub-ecosystems of those compatible ecosystems and not the other way around. You need to modify the comptaible chain tomls and add "Vanity ETH Generator Poly" into each sub-ecosystem list

data/ecosystems/v/Vanity-ETH-Generator-Poly
```diff
@@ -0,0 +1,25 @@
+# Ecosystem Level Information
+title = "Vanity ETH Generator Poly"
+
+sub_ecosystems = []
+
+github_organizations = ["https://github.com/samaeleth"]
+
+# Repositories
+[[repo]]
+url = "https://github.com/samaeleth/vanity-eth-generator/blob/master/Poly-Vanity-ETH"
```
jubos, line 10: These aren't repos.  You should just use url = "https://github.com/samaeleth/vanity-eth-generator"  one time.

jubos (changes requested):

PR #813, opened 2023-12-22
Added Mellow Protocol

armoking32:


eherrerosj:
We have an ecosystem for Mellow Protocol, I've internally added it a sub-eco of Polygon zkEVM and it will also rollup to Polygon. You'll see the updates when we sync the repo. Thanks!

PR #822, opened 2023-12-22
add zkSwap Finance

ZkSwapFinance:


data/ecosystems/z/zkswap-finance.toml
```diff
@@ -0,0 +1,7 @@
+# Ecosystem Level Information
+title = "zkSwap Finance"
+
+sub_ecosystems = []
+
+github_organizations = ["https://github.com/ZkSwapFinance"]
```
eherrerosj, line 6: this is not a github org url, it's a user url

eherrerosj (changes requested):
almost there. We can't scrape user profiles, only orgs, could you instead add the 3 individual repos from the user into the repo list below?

data/ecosystems/z/zkswap-finance.toml
ZkSwapFinance: Thank you, we have revised to 3 repositories.

eherrerosj (approved):

PR #823, opened 2023-12-22
add Fibrous Finance

0xKermo:


eherrerosj (changes requested):
You need to create a new ecosystem for Fibrous Finance. For that, you need to create a new toml file like fibrous-finance.toml with ttitle "Fibrous Finance" and add the org in it (https://github.com/Fibrous-Finance)

0xKermo:
I updated pr to just add fibrous finance repos

PR #827, opened 2023-12-22
Update zksync.toml with Vest Exchange

vest-ren:


data/ecosystems/z/zksync.toml
```diff
@@ -4811,6 +4816,3 @@ url = "https://github.com/zy19900820/ethers"
 
 [[repo]]
 url = "https://github.com/zycta/zycta-dapp"
-
-[[repo]]
-url = "https://github.com/dappsheriff/zksync-contracts"
```
eherrerosj, line 4816: we prefer to keep archived / missing repos because we have a persisted history that we can attribute to the parent ecosystem, please revert this deletion

data/ecosystems/z/zksync.toml
```diff
@@ -4274,6 +4275,10 @@ missing = true
 [[repo]]
 url = "https://github.com/vertekfi/hardhat-template"
 
+[[repo]]
+url = "https://github.com/VestLabs/papers/blob/main/zkRisk/zkRisk.pdf"
```
eherrerosj, line 4279: this is not a correct repo url

data/ecosystems/z/zksync.toml
```diff
@@ -25,6 +25,7 @@ sub_ecosystems = [
   "SyncSwap",
   "TrueFi Engineering (Ethworks)",
   "Uniswap",
+  "Vest Exchange",
```
eherrerosj, line 28: You need to create a new ecosystem for "Vest Exchange". For that, you need to create a new toml file like vest-exchange.toml with title "Vest Exchange" and add the org in it (https://github.com/VestLabs). This is assuming VestLabs dedicates most of it efforts to work on a projects(s) in ZKSync. Otherwise, just place here the specific repos that contribute to ZKSync

eherrerosj (changes requested):
left a few comments

jubos (changes requested):
Please see enrique's comments above

jubos:
Going to close this one out.  Please reopen when you can address the requested changes.

PR #835, opened 2023-12-22
Add solana-token-list.toml

JuanRdBO:


eherrerosj:
Repo already part of Solana

PR #847, opened 2023-12-22
Update starknet.toml

codekaya:
added early_starkers project on starknet
added dojo_arena project on starknet


eherrerosj (approved):
awesome thx

PR #849, opened 2023-12-22
Add Diode Ecosystem

themacexpert:
Adds Diode Ecosystem. They previously had their own chain but they are building exclusively on Moonbeam now.


eherrerosj (approved):
fantastic thx!

PR #857, opened 2023-12-22
Added Kleo Network 

vaibhavgeek:
Kleo Network is web3 social on top of polygon and solana! Learn more about Kleo Network - https://www.kleo.network/

### DIFF
--- a/data/ecosystems/p/polygon.toml
+++ b/data/ecosystems/p/polygon.toml
@@ -248,6 +248,7 @@ sub_ecosystems = [
   "Kirobo",
   "Kitty Party Finance",
   "Kleros",
+  "Kleo Network",
   "KlimaDAO",
   "KogeCoin",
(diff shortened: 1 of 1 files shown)


eherrerosj (changes requested):
Some feedback:
1. You added an org URL, not a repo URL
2. You added an ecosystem for which a toml file doesn't exist. You need to create a new ecosystem for Kleo Network. For that, you need to create a new toml file like kleo-network.toml with title "Kleo Network" and add the org in it (https://github.com/Kleo-Network)

jubos:
Added a Kleo Network sub ecosystem and manually merged your commit here: [9f4be16](https://github.com/electric-capital/crypto-ecosystems/commit/9f4be16afd31183be800e8855a26c785b9a57a26)

Thank you for the change suggestions.

PR #858, opened 2023-12-22
Update near.toml - Added ShardDog related repos

joe-rlo:


eherrerosj (changes requested):
You need to create a new ecosystem for Ready Layer One. For that, you need to create a new toml file and add the corresponding org / repos to it.

Then , you can remove the repos you added from the NEAR file

jubos (changes requested):
See Enrique's comments above.

jubos:
Looks like this got added by @s-n-park.  I am going to close it out.

PR #864, opened 2023-12-22
Update hyperlane.toml to include ecosystem

avious00:


avious00:
thanks in advance @eherrerosj

PR #870, opened 2023-12-23
Create polygon-miden.toml

gubloon:
Add repo details for Polygon Miden


eherrerosj (changes requested):
Is this compatible with Polygon and/or EVM and/or Ethereum? If so, "Polygon Miden" should be added to the sub_ecosystems list of the corresponding ecosystems

PR #808, opened 2023-12-22
Create issuaa.toml

avocadodefi:
Create issuaa.toml

PR #877, opened 2023-12-23
Create ISSUAA.toml

avocadodefi:
Create ISSUAA.toml


eherrerosj (changes requested):
Like I said in my previous PR, "ISSUAA" should be in the list of sub_ecosystems of the compatible chains. id, you'll need to modify the ethereum, base, zksync erc, etc tomls

eherrerosj:
duplicate of https://github.com/electric-capital/crypto-ecosystems/pull/808

PR #884, opened 2023-12-23
Create amunswap.toml

AmunRaCode:


eherrerosj:
closing because https://github.com/AmunRaCode is a user ant not an org. Please add the corresponding repos to the toml, remove the current org link and add the ecosystem as a sub-ecosystem to the compatible chains if any

PR #885, opened 2023-12-23
Create mantle0x.toml

mantle0x:


eherrerosj (changes requested):
some feedback:
1. remove the org link since its a user link and not an org link
2. add "Mantle0x to the sub_ecosystems list of the compatible chains / ecosystems

eherrerosj:
can't see how Mantle0x would be an ecosystem since core repo has no relevant code (?)

PR #897, opened 2023-12-23
add/update-klaytn-ecosystem

ayo-klaytn:


eherrerosj (changes requested):
Some fixes needed but this will be a great addition. [Here's the CI validation](https://github.com/electric-capital/crypto-ecosystems/actions/runs/7307328718/job/19912933194?pr=897#step:5:55) that runs automatically to do some verifications. As you can see, in order for sub-ecosystems to be added, a corresponding toml file with the same title as the sub-ecosystem added must exist. Looks like a lot of the sub-ecosystems don't have a toml associated and you'll need to create it (check the README for more details), or let me know if it is not clear.

You think you can have it ready by tomorrow Sunday (hard deadline)?

PR #898, opened 2023-12-23
Add jaguard2021/zkSync-Greeter-Contract into zksync.toml

jaguard2021:


jaguard2021:
Awesome, thanks for your paying atention

PR #914, opened 2023-12-23
Add and Organize edenia projects

xavier506:
- added libre ecosystem 
- reorganized edenia projects under corresponding ecosystems
- added tikuna project - an ethereum foundation grant


eherrerosj (changes requested):
looking nice. Just one thing, could you add Libre as a sub_ecosystem of the corresponding compatible ecosystems / chains?

xavier506:
thanks for the feedback!, [libre](https://libre.org) is its own L1 chain based on the antelope protocol, other chains based on the same protocol such as eos, wax, telos , proton have done the same by creating their own ecosystem file. I am not sure what  ecosystem it would go under.

PR #921, opened 2023-12-23
Update starknet.toml

barbayrak:
Official Starknet Amsterdam Hackathon project. Starkyield is a yield aggregator for starknet that communicates with Dex and Lending protocols like Jediswap , ZkLend , XBank.


eherrerosj:
duplicate repo

PR #924, opened 2023-12-23
Add cipher example

SriCrypt:
Tried creating a function to cipher string inputs in cairo


eherrerosj (changes requested):
https://github.com/SriCrypt is a user link not an org link

PR #934, opened 2023-12-24
Update starknet.toml

ArcticDEX:
ArticDEX added


eherrerosj (changes requested):
I don't think ArticDEX exists as an ecosystem (ecosystems have its own toml file), please create an ecosystem for it

eherrerosj:
done in a different PR

PR #938, opened 2023-12-24
Create CrossWeave-Protocol.toml

NestorLtd:
added


eherrerosj (changes requested):
please add "CrossWeave Protocol" to the list of sub_ecosystems of the corresponding compatible blockchains

PR #959, opened 2023-12-24
Create amunswap.toml

AmunRaCode:
Amun Swap created.


eherrerosj (changes requested):
please add in this PR "Amun Swap" to the list of sub_ecosystems of the compatible blockchains (each has its own toml that you'll need to modify)

eherrerosj:
repo has no relevant code that can be attributed to other blockchains currently. Can't be considered ecosystem because the owner is not an org

PR #961, opened 2023-12-24
Create kommunitas.toml

kommunitas:


eherrerosj (changes requested):
please add, in this PR, "Kommunitas" to the list of sub_ecosystems of the compatible blockchains (each has its own toml that you'll need to modify)

kommunitas:
@eherrerosj done

jubos:
@kommunitas, you shouldn't add Kommunitas as a sub ecosystem for itself as that will create a loop.  @eherrerosj was suggesting adding it to Polygon as your website would suggest.

kommunitas:
@jubos @eherrerosj 
I have fixed it, please check

eherrerosj (approved):
looks good now, thanks @kommunitas

PR #1022, opened 2023-12-28
Update hedera.toml to include pixfi

hhe44:


eherrerosj:
Since it claims to be in 12 chains, could you add PixFi to the other 11 chains please?

hhe44:
> Since it claims to be in 12 chains, could you add PixFi to the other 11 chains please?

That's a fair point; I've added PixFi to the other 11 chains in my most recent commit.

Please let me know if there's anything else! 👍

PR #1053, opened 2023-12-31
Update easidao.toml

hypn0t1z:


data/ecosystems/e/easidao.toml
```diff
@@ -3,59 +3,17 @@ title = "EasiDAO"
...
-url = "https://github.com/EaseDeFi/snapshot-spaces"
-
-[[repo]]
-url = "https://github.com/EaseDeFi/synthetix-faq-bot"
-
-[[repo]]
-url = "https://github.com/EaseDeFi/yield-server"
-
-[[repo]]
-url = "https://github.com/EaseDeFi/yinsure-finance"
```
eherrerosj, line 61: why are you proposing to remove this repos?

eherrerosj (changes requested):
we usually keep all repos including 404s, public repos that became private, forks, etc because we have ways to handle them

jubos (changes requested):
Please don't remove the existing repos as we keep them for archival support.

data/ecosystems/e/easidao.toml
hypn0t1z: I already updated, thank you

hypn0t1z:
> we usually keep all repos including 404s, public repos that became private, forks, etc because we have ways to handle them

hypn0t1z:
> Please don't remove the existing repos as we keep them for archival support.

Thank you for your feedback, I already updated

PR #1076, opened 2024-01-13
:memo: Update repos and add Magicblock

GabrielePicco:
# :memo: Update repos and add Magicblock


jubos:
Thanks for this, lgtm.

PR #1079, opened 2024-01-16
Add Ledger to Ethereum sub-ecosystems

alexandremgo:
Updating `data/ecosystems/e/ethereum.toml` to add Ledger as a sub-ecosystem

Thx !


jubos:
Thank you for the PR.  I am going to close this one.  We put wallets and infrastructure that support a broad number of ecosystems as their own top level ecos instead of as sub ecosystems of any particular ecosystems.

PR #1104, opened 2024-01-24
Update aleo.toml

miazn:
Adding Aleo repositories found through a custom github indexer


jubos:
Thanks, this is really helpful!

PR #1110, opened 2024-02-05
update icp ecosystem

letmejustputthishere:


data/ecosystems/i/internet-computer.toml
```diff
@@ -145,15 +143,29 @@ sub_ecosystems = [
...
   "Zondax",
+  "CrowdGov",
+  "Helios Labs",
+  "Helix-ex",
+  "Internet Computer",
+  "RankRadarICP",
+  "Tigris-Trade",
+  "Ledger",
+  "DFINITY Sandbox",
+  "Octopus Network",
```
jubos, line 154: Ledger and Octopus Network shouldn't be part of this as Ledger is a more general wallet and Octopus Network is a NEAR restaking protocol.

jubos (changes requested):

data/ecosystems/i/internet-computer.toml
letmejustputthishere: thanks for the heads up and nice to meet you. octopus is actively building on the IC right now so it should be part of this list. i will remove ledger though

letmejustputthishere: see here @jubos 
https://docs.google.com/document/d/10RSrGfj2Z-LmVhuAXX0Jon-9P9CO6k5ZNfS2hrQB72I/edit

jubos:
Thanks for the changes!

PR #1116, opened 2024-02-14
Supplement Noir Lang as sub ecosystems of Ethereum and Scroll

Savio-Sou:
Noir is a programming language for writing Zero-knowledge programs. It supports [generation of Solidity verifiers](https://noir-lang.org/docs/how_to/how-to-solidity-verifier) for ZK programs written in it, extends ZK workflows into Ethereum and EVM ecosystems and enable ZK applications in such blockchain environments.

References:
- [Ethereum storage proofs in Noir](https://github.com/aragonzkresearch/nouns-anonymous-voting)
- [Scroll's Noir workshop](https://www.youtube.com/watch?v=xLQfMu-shYA)

### DIFF
--- a/data/ecosystems/s/scroll.toml
+++ b/data/ecosystems/s/scroll.toml
@@ -6,6 +6,7 @@ sub_ecosystems = [
   "Eigen Network",
   "ISSUAA",
   "LayerZero",
+  "Noir Lang",
   "Pimlico",
   "Sushi Swap",
   "SyncSwap",


jubos:
Thank you for flagging this. 

I think Noir is more general like Move or Solidity, so shouldn't be a direct sub ecosystem to any crypto ecosystem in particular.

I will likely remove it from Aztec in the future at some point since it is looking to outgrow that particular ecosystem.  It did start with Aztec Labs though, so will keep it there for now.

PR #1119, opened 2024-02-14
Add more to the Stellar ecosystem.

ElliotFriend:
Adding lots of projects, repos, and organizations to the Stellar ecosystem.

It seems that (generally) these toml files are kept in alphabetical order, so I've tried to keep that convention going. If that's incorrect, I'm happy to make any changes necessary.


ElliotFriend:
Had a bunch more to add this afternoon, but now I'm done (for sure!)

ElliotFriend:
Hi, @jubos! Sorry to bug you, but wanted to ask for a review of this PR. Are there any changes you'd like to see? Or, is it good as-is?

ElliotFriend:
Hey, there, @jubos! Thanks for the feedback.

Yeah, I can make a PR that only includes the "obvious" ones. I'll take a look over all the additions, and give a second thought. Are there specific additions here you can share that might not fit with the Stellar ecosystem?

Maybe I've been thinking about it wrong, too. In my head, if a project/repo was "building on" Stellar, then I added it to the list. I was even considering that to be true in the event that they were incorporating Stellar as a secondary/tertiary/etc. chain. Perhaps that was over-zealous? Is the definition of "attributable" more about which chain the project _originates_ on? I may have gotten confused with the recent EC updates to include multi-chain devs/projects.

Thanks again!

ElliotFriend:
I took a first stab at reverting any "questionable" additions. I took most of these repos/orgs from some of our internal grant applications. Maybe some are not as far along in the process of integrating Stellar, which would likely mean they're not _yet_ right to include in our ecosystem?

It might also help to know that "Soroban" is the name of the smart contracts platform that was build for inclusion within the Stellar network (not as an L2, but has been included in the core network now). Anywhere you see "soroban," you can pretty much substitute the word Stellar, tbh.

Let me know if this covers all the "debatable" additions. Or, if there are still some more that aren't as clear-cut as you'd like. Thanks!

ElliotFriend:
@jubos Sorry to bug you again. Any suggestions or changes you'd like to see here?

data/ecosystems/s/stellar.toml
```diff
@@ -18,23 +18,40 @@ sub_ecosystems = [
...
   "https://github.com/aurora-rs",
   "https://github.com/blend-capital",
   "https://github.com/blocktimefinancial",
+  "https://github.com/bp-ventures",
+  "https://github.com/CometDEX",
   "https://github.com/CommuniDAO",
   "https://github.com/Creit-Tech",
   "https://github.com/fairxio",
+  "https://github.com/fluxitystellar",
+  "https://github.com/GladiusClub",
```
jubos, line 31: Let's move these github orgs to sub ecosystems since it is more flexible in the future especially if the orgs work in multiple ecosystems.

jubos (changes requested):
Thank you for these additions.  Overall look good, but let's move the orgs to specific sub ecosystems and then add them to Stellar.  This allows for more flexibility.

data/ecosystems/s/stellar.toml
ElliotFriend: @jubos I can do that, for sure! Couple questions first:

- Are there some "guidelines" you all like use to determine if something should be a sub-ecosystem, as opposed to just a `github_organization`?
- Should I move _all_ of the github orgs I've got in this PR, or just some of them?
- More generally speaking, does EC prefer to have sub-ecosystems set up rather than github orgs?

ElliotFriend:
@jubos I've gone ahead and changed _most_ of the added github orgs to distinct sub-ecosystems. I left some of the github orgs that are particularly Stellar-specific, and it doesn't seem (to me at least) that there will be a high likelihood of becoming multi-chain orgs. Of course, I'm more than happy to change those to sub-ecosystems, too, if you'd like.

Thanks!

jubos:
Merged, thanks for all of the adjustments.

PR #1122, opened 2024-02-15
Add Solmate

zerosnacks:
Adds [Solmate](https://github.com/transmissions11/solmate), a popular Solidity library.


zerosnacks:
Appears to already be included

PR #1129, opened 2024-02-19
Update fuel.toml   Add Sway-Amplify-Multiplier

jaguard2021:
Amplify-Multiplier is a project that encompasses various interconnected components to support the development and integration testing between Rust applications and the Sway framework


Boss4855 (commented):
show-filters-no-rgb

Boss4855:
> show-filters-no-rgb

PR #1131, opened 2024-02-20
Add DefiLlama to Ethereum

0xn4de:
This PR adds the DefiLlama ecosystem to ethereum.toml as a sub-ecosystem of Ethereum.


jubos:
DefiLlama is a general ecosystem, so we don't add it as sub ecosystem to any layer 1.

PR #1143, opened 2024-02-28
Added Fennel Protocol

Romulus10:
Fennel is a Web3 communications hub for medical, humanitarian, and emergency services technologies.


jubos:
Thanks!

PR #1153, opened 2024-03-05
Add Data Always open mev-boost data

dataalways:
This PR adds a CC0 daily updated dataset used to track Ethereum timing games and the relay ecosystem.


jubos:
Thanks!

PR #1165, opened 2024-03-10
Added Celenium Ecosystem

vvuwei:
Added ecosystem level information for [Celenium.io](https://celenium.io) - Blockchain Explorer for Celestia and based on it networks

Celenium.io is now a blockchain explorer that provides not just information about the Celestia network, but also insights into the broader ecosystem built around Celestia. This means you can use Celenium.io to explore not only the Celestia blockchain itself, but also the activity and data of other networks that are built on top of Celestia.

### DIFF
--- a/data/ecosystems/a/aevo.toml
+++ b/data/ecosystems/a/aevo.toml
@@ -1,7 +1,9 @@
 # Ecosystem Level Information
 title = "Aevo"
 
-sub_ecosystems = []
+sub_ecosystems = [
+  "Celenium",
+]
 
 github_organizations = ["https://github.com/aevoxyz"]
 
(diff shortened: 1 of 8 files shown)


jubos (changes requested):
Thanks for this.  

I would start with creating the ecosystem and adding it to Celestia only as a sub ecosystem.  If you start to add more ecosystem support in the repos or on the website, adding it as sub ecosystems to the others later would make sense.

vvuwei:
👋 Hey!

Thank you for your feedback. Are you suggesting we remove Celenium as a sub-ecosystem from Lyra, Aevo, Manta, Orderly, and the Public Goods Network? I'd like to note that these networks are supported within Celenium via our rollups leaderboard, which you can view here: https://celenium.io/rollups.

jubos:
Thanks for the clarification.  I will add it.

PR #1166, opened 2024-03-11
Add web-bip39 to general.toml

hujiulong:
Add [web-bip39](https://github.com/hujiulong/web-bip39) to general.toml


jubos:
Thanks!

PR #1171, opened 2024-03-12
Add kevinweb3 repo to ecosystems

kevinweb3:


kevinweb3:
ok

data/ecosystems/w/web3toolbox.toml
```diff
@@ -0,0 +1,34 @@
+# Ecosystem Level Information
+title = "kevinweb3"
+
+sub_ecosystems = []
+
+github_organizations = ["https://github.com/kevinweb3"]
```
jubos, line 6: This is a github user, not an organization.  Can you remove this?

Separately, given it's a user and not an ecosystem, it's preferable to add these single repos to their respective ecosystems.

jubos (changes requested):

data/ecosystems/w/web3toolbox.toml
kevinweb3: okay, understand! I will change it, thanks!

PR #1182, opened 2024-03-15
Update aptos.toml

zacharyr0th:
updated repo list

### DIFF
--- a/data/ecosystems/a/aptos.toml
+++ b/data/ecosystems/a/aptos.toml
@@ -12,14 +12,11 @@ sub_ecosystems = [
   "Aptos Rabbit",
   "Aptosis",
   "AptStats",
-  "Aries Markets",
   "AUX Exchange",
   "BlockEden.xyz",
   "BlockPi Network",
   "Bware Labs",
   "Chainbase",
-  "ComingChat",
-  "Deploy DAO",
   "Ditto Finance",
   "Econia Labs",
   "Enchanter Finance",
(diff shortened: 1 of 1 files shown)


jubos (changes requested):
Thanks for the first pass here.

This diff is too large to be reviewed and it also pulls in too many non Aptos repos.  If you are using automation to generate the list, I would suggest breaking this up into a few sub PRs that are easier to review.  

For instance, for all of the organizations that you find building in Aptos, add a sub ecosystem for each one, hook them up in the aptos.toml file, and make smaller PRs so they are easier to review.  The ICP and Moonbeam ecosystem teams have done a good job of this as examples.

Separately, any repo that doesn't mention Aptos at all should be done in a separate PR so that we can have a discussion on whether it resides in the Aptos ecosystem.  The diff removes a lot of sub ecosystem that do appear to be in Aptos ecosystem.  Was that your intention?

Lastly, if you can sort your edits of the toml files with your automation tool, it makes the diff much cleaner to read since it doesn't have so many deletions and additions.

zacharyr0th:
The automation was pulling the [dependents](https://github.com/aptos-labs/aptos-core/network/dependents?package_id=UGFja2FnZS0zMDYzNDc5NjM0
) on the most popular [Aptos-Core ](https://github.com/aptos-labs/aptos-core) related packages and wasn't intended to remove active repos. 

So I should essentially add in sub-ecosystem for each of the projects on [this](https://aptosfoundation.org/ecosystem/projects/all) page? Some of them are closed source still but I noticed that some of the sub-ecosystems on moonbeam's .toml are not included as repos but just include the name in the sub-ecosystem section, such as [1beam](https://github.com/electric-capital/crypto-ecosystems/blob/d16797012c6355aba6c86c5f613a5d637df193a3/data/ecosystems/m/moonbeam.toml#L5).

PR #1190, opened 2024-03-20
Add Walnut to Starknet Ecosystem

jaipaljadeja:
Added [Walnut](https://github.com/walnuthq) to Starknet Ecosystem

### DIFF
--- a/data/ecosystems/s/starknet.toml
+++ b/data/ecosystems/s/starknet.toml
@@ -54,6 +54,7 @@ github_organizations = [
   "https://github.com/starknet-io",
   "https://github.com/StarknetAstro",
   "https://github.com/starkware-libs",
+  "https://github.com/walnuthq",
   "https://github.com/z-korp",
 ]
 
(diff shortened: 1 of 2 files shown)


mazurroman:
Closes https://github.com/walnuthq/cairovm.codes/issues/93

data/ecosystems/s/starknet.toml
```diff
@@ -54,6 +54,7 @@ github_organizations = [
   "https://github.com/starknet-io",
   "https://github.com/StarknetAstro",
   "https://github.com/starkware-libs",
+  "https://github.com/walnuthq",
```
jubos, line 57: Since walnut is an org and has a dedicated website / twitter account, it's preferable to make an ecosystem and then attach it as a sub ecosystem of Starknet.  This gives it more flexibility in the future in case Walnut adds more ecosystems that they support.

jubos (changes requested):

jaipaljadeja:
@jubos thanks for the review. added walnut as sub-ecosystem in starknet.

xuhongbo (commented):
What is this change?

jaipaljadeja:
> What is this change?

@xuhongbo earlier i just added walnut in github organization of starknet, then @jubos mentioned some points about walnut and suggested that we add walnut as a sub-ecosystem itself. So made changes accordingly 🙌

jubos:
Thanks for the change!

PR #1204, opened 2024-03-27
Create langchain.toml

agodbug:


jubos:
Thanks for the PR, but Langchain isn't really in scope for crypto-ecosystems...yet.

PR #1212, opened 2024-03-28
Update eidos.toml

dockertest4paul:


jubos:
No need for this change, thx.

PR #1215, opened 2024-03-28
Add eth-duties to ethereum-consensus ecosystem

TobiWo:
## Summary

Eth-duties is a tool to track validator duties in order to help you decide when to have maintenance periods for your infrastructure or your deployments so that you do not miss any important/big rewards.


jubos:
Thanks!

PR #1237, opened 2024-04-04
add debugging lib for chain impl by golang

cch123:
holmes is used to debugging some blockchain impl by golang, including https://github.com/okx/exchain, https://github.com/okx/okbchain


jubos:
Thanks for the PR, but this tool is a general tool for golang debugging, so it's out of scope for this taxonomy.

PR #1241, opened 2024-04-11
Correcting Apillon GH Org

pavlamijic:
"https://github.com/Apillon-web3" gives 404 and I think those repos can be entirely removed. New ones under "https://github.com/Apillon" should be pulled.


jubos:
Manually merged here: [c7781c6](https://github.com/electric-capital/crypto-ecosystems/commit/c7781c6a8523c483b81915d78e9bd7e23918de35).

Thanks!

PR #1261, opened 2024-05-08
Update near.toml

PiVortex:
Add roketo org


jubos:
Thanks for the PR.  Roketo is already in the NEAR ecosystem as a subecosystem so it's covered.

PR #1270, opened 2024-05-10
update icp ecosystem

letmejustputthishere:
@jubos some new updates :)


jubos (changes requested):
Thanks for this, can you help with sorting these so the validation passes.

letmejustputthishere:
> Thanks for this, can you help with sorting these so the validation passes.

@jubos is your tooling for sorting available somewhere? i'd really like to avoid replicating the logic due to time constraints

jubos:
Merged, thank you for the fixes.

PR #1277, opened 2024-05-26
add aicrypto group

ThanThoai:


jubos:
Hi, these don't look like active repos.  Are you sure you want to add this?

ThanThoai:
Yes, I want to add this

jubos:
Thanks, this doesn't look crypto related.  Let's add when there is some clear crypto activity.

PR #1279, opened 2024-05-30
Update IoTeX Ecosystem

GiupiDeLuca:


data/ecosystems/a/ator.toml
```diff
@@ -0,0 +1,57 @@
+# Ecosystem Level Information
+title = "Ator"
+
+sub_ecosystems = []
+
+github_organizations = [
+"https://github.com/iopay/iopay-integration-example"
+]
```
jubos, line 8: This should be a repo and not an organization

data/ecosystems/i/iotex.toml
```diff
@@ -201,18 +168,6 @@ url = "https://github.com/iotexproject/iips"
...
-missing = true
-
-[[repo]]
-url = "https://github.com/iotexproject/iopay-extension"
-missing = true
-
-[[repo]]
-url = "https://github.com/iotexproject/iopay-extension-wallet"
-missing = true
-
```
jubos, line 215: Please keep the missing attributes since we use these for historical reasons.

jubos (changes requested):
Overall looks good, just a few small tweaks and we can merge it.

data/ecosystems/a/ator.toml
GiupiDeLuca: Sorry, my bad, this was actually a typo, the array should be empty - I fixed it

data/ecosystems/i/iotex.toml
GiupiDeLuca: @jubos Just to make sure I understand: I removed these since they were not relevant. So do you want me to add them back with the "missing" attributes?

jubos: Yes, because they existed at one time in case.

GiupiDeLuca: @jubos thank you! All set on my end!

jubos:
Merged, thank you for the improvements.

PR #1292, opened 2024-06-07
Update near.toml

s-n-park:
Adding additional individual repos from DevHub hackathon submissions and manually reviewing GH code search.


jubos (changes requested):
There are a few invalid repos in here.  Can you please clean these up?

s-n-park:
@jubos - thanks for the feedback, went back through the list and cleaned up the suffix

jubos:
Thanks for the adjustment!

PR #1293, opened 2024-06-10
Update movement.toml

0xNeit:


jubos (changes requested):
These need url = "" around then to be valid.  Separately, can you move the razorlabsorg into its own ecosystem and then add to Movement?

jubos:
Manually merged this here: [8f341e3](https://github.com/electric-capital/crypto-ecosystems/commit/8f341e36449f8ad2a5832126ae7d2ad3ec615c26)

PR #1327, opened 2024-06-19
Add types-js to  starknet.toml

tabaktoni:
[repo](https://github.com/starknet-io/types-js )
[npm](https://www.npmjs.com/package/@starknet-io/types-js)


jubos:
Thank you

PR #1330, opened 2024-06-19
[chore] Adds missing Argent data to starknet.toml

graemeblackwood:
Adds missing argentlabs Github organisation and corrects Dappland repo to canonical repo in argentlabs Github org.


jubos:
Going to close this out as Argent is already a sub ecosystem of Starknet.

PR #1340, opened 2024-06-19
Updated : Added repos related to Karnot and Madara

ocdbytes:
- Added Madara repos.
- Added Karnot repos.
- Added some other repos from Moonsong Labs related to madara.
- Added one repo from abdelstark related to madara


jubos:
Thanks!

PR #1367, opened 2024-06-21
feat: add starknet-abi package

LordGhostX:
Adds the Starknet ABI package for encoding and decoding Starknet contract calls and events: https://github.com/NethermindEth/starknet-abi


jubos:
Thanks

PR #1403, opened 2024-06-26
Update sub-ecosystems and add new repos for Base

wbnns:
This PR updates the previous metadata for Base, adding new sub-ecosystems and repos.

### DIFF
--- a/data/ecosystems/b/base.toml
+++ b/data/ecosystems/b/base.toml
@@ -1,71 +1,614 @@
-# Ecosystem Level Information
 title = "Base"
 
 sub_ecosystems = [
-  "Aave",
-  "Aerodrome",
-  "API3",
-  "Aura Finance",
-  "Balancer",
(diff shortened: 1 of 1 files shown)


jubos (changes requested):
This is a good start.  I would suggest breaking it up into a few PRs.  First start with adding the existing sub ecosystems to Base.  Then add these new sub ecosystem tomls for the brand new ecosystems that don't exist, so it passes the validation.

jubos (changes requested):
This is getting closer.  I would suggest removing the non-existent ecosystems and sort of the repos for the first pass, and then we can add new ecosystems in separate PRs.

wbnns:
@jubos 

Heya Curtis! Thanks for the feedback! This should all be resolved and ready for merging. 👍

jubos:
Thank you for the updates!

PR #1417, opened 2024-07-01
Add Cheesecake Labs as a sub-ecosystem of Stellar

ElliotFriend:
Cheesecake Labs is a very active dev organization in the Stellar ecosystem. This commit adds them as their own sub-ecosystem of Stellar. It seemed like a sub-ecosystem was warranted for two reasons:

1. (Mainly this one) They've developed their own `stellar-plus` SDK which wraps/enhances the SDF-maintained JavaScript SDK. There are several repositories using that as a dependency (as can be seen at the bottom of the [new] `cheesecake-labs.toml` file).
2. There's quite a long list of repos in their GH org.

If there's a better rubric you all use for determining if something is more fit to be in the `sub_ecosystems` vs the `github_organizations`, I'm more than happy to adjust this (and any following) PRs accordingly.

Thanks!

> _Note:_ I'm well aware that we're past the deadline for the midyear update, and I have no expectations that this change would be included/reflected in that midyear data. Just trying to keep things up-to-date moving forward.

### DIFF
--- a/data/ecosystems/c/cheesecake-labs.toml
+++ b/data/ecosystems/c/cheesecake-labs.toml
@@ -1,0 +1,274 @@
+# Ecosystem Level Information
+title = "Cheesecake Labs"
+
+sub_ecosystems = []
+
+github_organizations = ["https://github.com/CheesecakeLabs"]
+
+# Repositories
+[[repo]]
+url = "https://github.com/CheesecakeLabs/.github"
+
+[[repo]]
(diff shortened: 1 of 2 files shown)


jubos:
Good addition, thanks!

jubos:
In general, if a github organization has its own website, discord, twitter handle, it warrants making it a dedicated ecosystem.

PR #1422, opened 2024-07-03
Update algorand.toml: Remove dupes

auditoshi:
Removed 4 dupes


jubos:
These repos look fine and aren't duplicates, so going to close this out.

PR #1430, opened 2024-07-11
Create apt-casino.toml

AmaanSayyad:


jubos:
This repo is already tracked in the main aptos toml.  We can upgrade it to an ecosystem once it has a community and discord, etc.

PR #1439, opened 2024-08-09
Update Lido Ecosystem

tamtamchik:
Hey there! Thanks for all your hard work! Your repo is amazing.

I want to tidy up all the repos for Lido and update the main protocol repo tags reference.

I'm a bit confused because the Readme says repo records shouldn't be added, but I've noticed several PRs adding repos to the projects. I'm unsure how to update the organization's records. @jubos, can you please elaborate on this?


mariashen:
Thank you for the cleanup @tamtamchik!

On your question about "repos shouldn't be added" -- basically if we are already tracking a GitHub org for an ecosystem, all the repos under that GitHub org is automatically tracked. That's why it's not necessary to add more repos from a Github org that's already part of an ecosystem. Does that answer your question?

mariashen (approved):
This PR is approved, but these two repos were mistakenly removed and we will be adding these back in:

1. https://github.com/saurabh27-/lido-frontend-template-main
2. https://github.com/brown-qs/avo-vault-frontend

If there is some mistake here, let us know!

jubos:
Thank you.  I manually merged this here: 89a995ce18b71dcecca64cf70ec269b0b251bcab.

We keep the missing repos for completeness since they have a missing attribute.

tamtamchik:
Thank you @jubos! 😍

PR #1447, opened 2024-08-28
Adding new project- Update aptos.toml

reetbatra:
- added our project called Chrono's Gambit built during Aptos Winter School, which is a prediction marketplace using LMSR algorithm to set prices and determine payouts based on participants' inputs.


jubos:
Thanks, manually merged here: b74c1dadbb0749e722eddbfa516d0ae58d6622a9

PR #1467, opened 2024-09-02
Update aptos.toml

ayosheisTHICC:


data/ecosystems/a/aptos.toml
```diff
@@ -8018,6 +8019,9 @@ url = "https://github.com/Saturn-Radius/aptos-vault"
 [[repo]]
 url = "https://github.com/Saturn-Radius/oracle-aptos"
 
+[repo]]
```
jdesanto, line 8022: An open [ needs to be added to remove the syntax error

PR #1476, opened 2024-09-13
Add Web3j

gtebrean:


gtebrean:
@jubos could you please take a look at this and let me know if it's anything which should be updated?

jubos:
I am going to close this one out since all of these are already covered in Hyperledger and web3j is in the Ethereum ecosystem.

PR #1487, opened 2024-10-01
update klaytn to kaia

ayo-klaytn:


jubos:
Thank you, manually merged with some fixes above.

PR #1488, opened 2024-10-02
Update algorand.toml: Remove repo spamming commits, which fakes developer activity

auditoshi:
Removed 1 repo spamming commits, which fakes developer activity.

- `algotables/algotables.github.io`
  - Spams one or more commits a minute with a Google Script. Usually the commits are [empty](https://github.com/algotables/algotables.github.io/commit/ff1e9e9f118c47811ab007f09d0102f88a96af9b).
  - Doesn't interact with the blockchain, or provide documentation either.
  - Doesn't provide developer tools as the README suggests.

image

This skews rankings on websites using the `crypto-ecosystems` dataset.

### Artemis.xyz

image

### Cryptometheus

image

### DIFF
--- a/data/ecosystems/a/algorand.toml
+++ b/data/ecosystems/a/algorand.toml
@@ -1848,9 +1848,6 @@ url = "https://github.com/AlgoRealm/algorealm-cli"
 url = "https://github.com/AlgoStation/AlgoPass"
 
 [[repo]]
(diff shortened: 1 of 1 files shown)


jubos:
Seems reasonable to remove.  This repo is flagged as a bot driven repo in our dataset internally, so we don't count as an active developer.  Yet another reason that a naive commit count is not a great metric.

PR #1489, opened 2024-10-02
Update Space and Time ecosystem

AlejandroLanaspa:


jubos:
Thank you for the new ecosystems and additional org for SxT.  For an open source project like Space and Time that is more horizontal, those ecosystems would not be sub ecosystems to SxT just because they use the code in one aspect of their protocol.

AlejandroLanaspa:
Thanks for the update @jubos ! So, for those ecosystems, the correct approach would be to send only the repo that is fully related to Space and Time, right?

PR #1490, opened 2024-10-02
fix: update axelar-network.toml

Olanetsoft:
Add more ecosystem projects


jubos:
Manually merged above!  I removed a few of the repos not directly associated with Axelar or ones that were missing.

PR #1495, opened 2024-10-10
Update cosmos-network.toml to include XION

ash-burnt:


jubos:
Thank you, manually merged.

PR #1498, opened 2024-10-11
Update various ecosystems

themacexpert:
Update various ecosystems


mariashen (approved):
Thank you for the PR!

For reference for the future, we automatically track any repo part of a GitHub org. This means you do not need to add in repos if the org is already part of the toml file.

jubos:
Thank you for this.  It's very helpful.  Merged above.

PR #1529, opened 2024-10-13
Update Cardano Ecosystem

gufmar:
added Cardano relevant repositories


mariashen (approved):
Thank you for the PR!

This PR is approved, but the following repos will be excluded because they do not seem to be related to Cardano:
- https://github.com/xray-network/cloudflare-worker-output-load-balancer
- https://github.com/justmert/arbitrum-python-sdk
- https://github.com/tolgayayci/sora
- https://github.com/tolgayayci/dfinity-dfx-gui
- https://github.com/tolgayayci/stylus-vscode-extension
- https://github.com/veralidity/SecureVault

Please submit another PR if you think this is a mistake. We will add the other repos to the Cardano taxonomy!

gufmar:
Hi Maria
Before I may submit another PR, I don't want to appear obsessive, and rather ask here first. 

> ...do not seem to be related to Cardano:
> 
> * https://github.com/xray-network/cloudflare-worker-output-load-balancer

This one actually is a proxy front-end for all kind of Cardano related [endpoints](https://github.com/xray-network/cloudflare-worker-output-load-balancer?tab=readme-ov-file#endpoints).  See also the sample [config file](https://github.com/xray-network/cloudflare-worker-output-load-balancer/blob/main/src/servers.conf.example.ts) 

For all others I agree, from a pure current view on these repos, and without the context of how and why these things are currently build for future Cardano interactions it's not possible to assigne them to Cardano.

jubos:
Thank you, I manually merged your PR above and left the cloudflare-worker-output-load-balancer repo in place because it is contributed by xray-network, an active Cardano builder.

PR #1533, opened 2024-10-14
Adding repositories related to CIP54 Cardano

kieransimkin:


jubos:
Thanks for this!  Good feedback, perhaps we can add autosort to the CI.

PR #1537, opened 2024-10-18
[WIP] Update from Ecosystem Data - Developer Report Updater (p_7NCMQz6)

emilyJLin95:


jubos (changes requested):
Thank you for this.  A few requests of improvements.

The sub ecosystem formatting is off so it's causing an unnecessary diff.

There are a lot of missing repos that 404 on visiting. 

Lastly, no need to change case on existing repos.  This creates an unnecessary diff.

PR #1542, opened 2024-10-23
Adding repositories directly related to SxT

AlejandroLanaspa:
Based on the feedback from [this PR](https://github.com/electric-capital/crypto-ecosystems/pull/1489) ,  I have added specific repositories that are built on Space and Time (SxT).


jubos:
Thanks, merged!

PR #1577, opened 2024-10-30
Added Blink generator repo to solana ecosystem

shubhiscoding:


mariashen:
Thank you for the PR! Error is due to unsorted repos. Otherwise looks good. Try 'make sort' to fix the issue

shubhiscoding:
> Thank you for the PR! Error is due to unsorted repos. Otherwise looks good. Try 'make sort' to fix the issue

thanks, fixed it!

PR #1579, opened 2024-10-30
Update IoTeX Ecosystem

GiupiDeLuca:


yizhao-ec:
Hello thanks for the PR, it seems like many of the repos added to seeed-studio are hardware focused and unclear how unrelated to crypto they are. Will close PR for now but please open new PR if you disagree.

PR #1585, opened 2024-10-30
Update bob.toml

derrekcoleman:
Add ecosystem projects as sub_ecosystems. Alpha sort.


yizhao-ec:
Hello thanks for the PR - however looks like most of these subecosystems will actually now be defined as multichain by our updated taxonomy and therefore will not directly connect to any specific chain. Will close this PR for now but feel free to re-open another PR in the future if you disagree.

PR #1604, opened 2024-10-31
add BoltWade Studio repo

doraemonidol:
Add the Brewmaster Game from BoltWade Studio


yizhao-ec:
Thank you for the PR! We will create a new ecosystem for BoltWade Studio (from #1601) and this repo will be attributed to starknet via the new subecosystem.

PR #1606, opened 2024-10-31
added Looop music

joeephwild:


yizhao-ec:
Hello thanks for the PR! From looking at the contracts, seems like this git organization should go under solana rather than starknet?

joeephwild:
No actually it’s on starknet the solana code was an old code written an
along time ago, we have a starknet code written in Cairo there.

You can check again 😁, thanks.

yizhao-ec:
Sounds good - thanks! I removed the duplicated attribution in both subecosystem and organization and created the subecosystem internally with the organization link.

PR #1607, opened 2024-10-31
Add BoltWade Brewmaster

noodthegd:


yizhao-ec:
Thanks for the PR, the repo should be merged already from #1601

PR #1613, opened 2024-10-31
Added hypersign protocol to the taxonomy

Vishwas1:
Requesting to add Hypersign protocol.


Vishwas1:
Unable to understand why validate action is failing. Anything needs to be done at our end ?

yizhao-ec:
No worries - I can fix the minor validation issues in batch. Thanks for the PR!

yizhao-ec:
Also, as FYI, after internal discussion we moved Hypersign Protocol from Cosmos Network to Cosmos SDK (which is atm only internal) but will attribute up to the Cosmos Network Stack ecosystem. Let me know if you disagree.

PR #1637, opened 2024-10-31
Add in Solana Repos to Solana.toml 

brimigs:


yizhao-ec:
Hello thanks for the PR! Just as FYI, after internal reviews we removed some repos that were added by this PR. See https://github.com/electric-capital/crypto-ecosystems/commit/05182223f6585a37da3c50ce69b758fa0a35ea1c for the details.

PR #1639, opened 2024-10-31
Batch_add Solana repos

Nelis-sol:


yizhao-ec:
Hello thanks for the PR! Just as FYI, after internal reviews we removed some repos that were added by this PR. See https://github.com/electric-capital/crypto-ecosystems/commit/4bf6e27b09580fdbe5cea24e9d1079b0e53468d6 for the details.

PR #1657, opened 2024-10-31
Update starknet.toml added grugs lair

bengineer42:


yizhao-ec:
Thanks for the PR! Another PR has already added these.

PR #1671, opened 2024-11-01
Update aptos.toml

zacharyr0th:


yizhao-ec:
Just as a FYI - after internal reviews we made some modifications after merging:
* Some sub_ecosystems were removed due to being multichain rather than aptos specific - see https://github.com/electric-capital/crypto-ecosystems/commit/642df992fdd3db5e5023c6cec2c4adede7120de6 for details (note the changes in repos in this commit are due to sorting)
* Some repos were removed due to not identified as aptos - see https://github.com/electric-capital/crypto-ecosystems/commit/d3f964905be5a86195f5dbeb213880adb0a78d5a for details, if you think some of these repos should not have been removed please let us know and open another PR

Thanks!

PR #1675, opened 2024-11-01
Added Collaborative Economics' Give-Stark

Utilitycoder:


yizhao-ec:
Hello, thanks for the PR! However looks like https://github.com/CollaborativeEconomics also contains many repos that are not related to starknet. Would recommend creating a new ecosystem for this git organization and then associate the chain specific repos to the chain ecosystems. Thanks!

PR #1692, opened 2024-11-05
Update bnb-chain.toml

STdevK:
added projects building on BNB Chain


yizhao-ec:
Removed `https://github.com/micro3dev` from github organizations as its just a github user and not able to be handled as an organization. Otherwise all added. Thanks!

PR #1743, opened 2024-12-18
Add WeaveVM

xylophonez:
[WeaveVM](https://wvm.dev) is a sovereign EVM network built towards the goal of solving the EVM storage dilemma with Arweave. It gives the coming generation of high-performance chains a place to settle and store onchain data, without worrying about cost, availability, or permanence.


jubos:
Thanks for this!

PR #1754, opened 2025-01-02
add "alpine-sshd" repo in RunOnFlux add a Customize wasp Add the Door for Pi  Submit IBM-Blockchain submit the SavjeeCoin

zhuweiDark:
add "alpine-sshd" repo in RunOnFlux

Run On Flux
Flux, Your Gateway to a Decentralized World
The simplest/smallest SSH Server
This is a very simple alpine based SSHD server, listening on port 22. It has no fancy stuff. Just SSHD service running in foreground. It allows only key based access, and the user is root. Useful as a jump host inside kubernetes clusters, to further connect to your other secure services behind firewall, etc. You can use it for port forwarding, ssh-agent forwarding, etc.

### DIFF
--- a/data/ecosystems/d/door.toml
+++ b/data/ecosystems/d/door.toml
@@ -1,0 +1,10 @@
+# Ecosystem Level Information
+title = "door"
+
+sub_ecosystems = []
+
+github_organizations = ["https://github.com/pi-apps/door"]
+
+# Repositories
+[[repo]]
+url = "https://github.com/pi-apps/door"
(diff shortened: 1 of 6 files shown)


jubos:
I am going to close this one out.  It is doing too many disparate changes in one commit.  Also, there is no need to add ecosystems like Ripple, which are already covered.  

Can you make new PRs with one idea / ecosystem per pull request?

zhuweiDark:
roger that， i will do it

PR #1762, opened 2025-01-09
Updated algorand.toml to add CompX Labs organisation

xxiled-plastic-cat:
Added CompX Labs org - https://github.com/compx-labs


jubos:
Thanks!

PR #1765, opened 2025-01-17
fix: improve directory creation error handling

crStiv:
- Add DirectoryCreationError variant to CEError enum
- Replace unwrap() with safe directory creation error handling
- Propagate directory creation errors properly through Result


jubos:
Thanks!

PR #1766, opened 2025-01-24
Update ton.toml

markysha:
Several steps were made to create this update:

1. Crawl top used dependencies of **ton** related repos
2. Validate and enrich list of top libs with https://docs.ton.org/develop/dapps/apis/sdk
3. For each dependency find repos that use them using **github code search** 
4. Filter out repos that already exist in [electric-capital/crypto-ecosystems](https://github.com/electric-capital/crypto-ecosystems) repo
5. Add new repos to **ton.toml** and pretty file to match file format

This code used for creating update https://colab.research.google.com/drive/1RuxC-QIeQHxAQlMOfsPkGjvMivg5bT1s#scrollTo=TayrgH-InUuq

Log:
```
parsed repos (can be repeats) 6714
skipped because org already exists 629
skipped because repo already exists 638
skipped because duplicate 3535
new repos 1912
```


jubos:
Thanks for this!  I appreciate the rust update as well.

PR #1779, opened 2025-02-23
Update aptos.toml

VanshKapoor07:


yizhao-ec:
Hello thanks for the PR! The repo looks good, however I don't see a new ecosystem created with name `CryptoAI` in this PR so that sub ecosystem is not valid. Could you remove it and I will approve PR once that is done

yizhao-ec (approved):

PR #1785, opened 2025-02-24
Added Decentralized events to the list

Saurav10codes:
Added Decentralized events to the list


yizhao-ec:
Hello, thanks for the PR. Which ecosystem (and which toml) is `Decentralized Events` referencing?

PR #1816, opened 2025-03-04
Update cardano.toml

zhekson1:
added p2p-defi project & repos


yizhao-ec:
Hello, thanks for the PR! The 2 repos looks great, however in order to add `P2P-DeFi` sub_ecosystem there needs to be a corresponding toml file for `P2P-DeFi`. If your goal is to only add those 2 repos, I would suggest removing the new sub_ecosystem instead.

yizhao-ec (approved):

PR #1817, opened 2025-03-04
Update cardano.toml

SamDelaney:


yizhao-ec:
Hello thanks for the PR. The repos looks great, however in order to add `Ikigai Technologies` as a sub_ecosystem, there need to be a corresponding toml created for `Ikigai Technologies`. In this case, the I may suggest just removing them from the sub_ecosystem instead.

SamDelaney:
@yizhao-ec done, and added a couple more repos I work on that I noticed weren't tracked :)

emrecaliskan:
Hey @SamDelaney this looks great, ty! 

Do you think you could sort the repos to fix the [failing CI/CD](https://github.com/electric-capital/crypto-ecosystems/actions/runs/13708521466/job/38339588170?pr=1817).

There is a `make sort` utility command available [here](https://github.com/electric-capital/crypto-ecosystems/blob/master/Makefile#L34C1-L34C5).

PR #1839, opened 2025-03-10
Update cardano.toml - add missing Cardano github organizations

Ryun1:
Hello!

I am adding 30 Github organisations which primarily focus on development within the Cardano ecosystem.

I will do a follow up pull request to add in other key repositories which either; sit under individuals or sit under organisations which do not primarily focus on Cardano.

thank you!


emrecaliskan:
Thanks for the PR! These look good to me. Ideally, some of these should be sub-ecosystems (e.g., https://github.com/masumi-network) of Cardano, as that would more accurately represent Cardano and its ecosystem. However, this is an acceptable middle ground to account for the developers but not for the sub-ecosystems.

PR #1843, opened 2025-03-13
Add IOTA as a subecosystem of Sui

tedks:
IOTA is a fork of Sui that contributes back changes to mainline Sui.

### DIFF
--- a/data/ecosystems/s/sui.toml
+++ b/data/ecosystems/s/sui.toml
@@ -20,6 +20,7 @@ sub_ecosystems = [
   "Hentis",
   "HyperPlay",
   "Interest Protocol",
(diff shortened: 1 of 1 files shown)


yizhao-ec:
Thanks for your PR. I believe IOTA might fit better as a sub-ecosystem under the [Move Stack](https://github.com/electric-capital/crypto-ecosystems/blob/master/data/ecosystems/m/move-stack.toml) rather than directly under Sui.

I'll close this PR for now, but feel free to open a new one if you agree with this suggestion or if you have specific reasons why IOTA should be categorized under Sui specifically.

Thanks again for your contribution!

PR #1847, opened 2025-03-14
Adding GoPlausible

emg110:
Adding GoPlausible to ecosystems and also under the Algorand ecosystem


emrecaliskan:
Thanks for the PR. Removing `AlgoPOAP` makes sense, but could you explain why you're removing individual repos (which seem to be yours) from Algorand? They still appear to be valid repos.

emrecaliskan:
That’s very thoughtful. No need to find the originals; we handle that in our backend. Could you please add them back for two reasons: 

1) We count original work in addition to forks. So, if you forked and worked on top of that, those are valid and should be counted. :)
2) They’re already in our taxonomy and backend, so it’s easier to just keep them.

Happy to merge once those changes are pushed. Thanks again!

emg110:
I will close this one and create a new one with just AlgoPOAP removal and GoPlausible additions @emrecaliskan

PR #1858, opened 2025-03-27
add SShift GPT project

spielcrypto:
Adding SShift GPT Project to aptos ecosystem

web app: https://sshiftgpt.com
gitbook: https://sshiftgpt.gitbook.io/sshiftgpt


yizhao-ec:
Hello thanks for the PR, in order to associate a new sub ecosystem for Aptos, `SShift GPT` needs to be its own ecosystem with its own toml file (and various attributes, eg. Github organization). In this case, you might want to just remove the `SShift GPT` from the sub ecosystem list.

PR #1865, opened 2025-03-28
Add moveduck

phuquiivo03:


yizhao-ec:
Hello, thanks for the PR. However it seems like the new organization link you provided in `moveduck.toml` is actually a repo link. Given that the new ecosystem only have 1 repo, it may make more sense to just have the repo link directly in Aptos. 

Please either fix the repo link under `github_organizations` and provide more evidence that it should be its own ecosystem, or move the repo link to under `aptos.toml` directly.

yizhao-ec:
Hello, thanks for making the updates! However, looking at https://github.com/0xDuckMove - its a Github user and not a Github organization account and therefore should be removed.

yizhao-ec (approved):

PR #1866, opened 2025-03-28
Add Cloud Web3 Agent project to aptos.toml

pisuthd:


yizhao-ec:
Hello, thanks for the PR. The new `Cloud Web3 Agent` ecosystem doesn't have its own corresponding toml file, and therefore can't be a child ecosystem. Please remove `Cloud Web3 Agent` from the sub ecosystem list for Aptos or create a new toml with evidence that `Cloud Web3 Agent` should be its own distinct ecosystem.

PR #1868, opened 2025-03-28
Adding Merkle AI Auto Trading  repo in aptos.toml

Suharaz:


yizhao-ec:
Hello thanks for the PR, since you added `Merkle AI Auto Trading` as a sub ecosystem for Aptos the structure require there to be a matching toml file for `Merkle AI Auto Trading` as well. Given that you're only looking to add the one repo, I think it may make sense to just remove `Merkle AI Auto Trading` from the sub ecosystem list.

PR #1870, opened 2025-03-28
Added Kaisen Assistant in aptos.toml

thecoderwithHat:


yizhao-ec:
Hello thanks for the PR, the link you provided as an organization "https://github.com/thecoderwithHat/Kaisen-Assistant" is actually a repo link instead of a Github organization. 

Also, it doesn't seem like there's any modification made to Aptos as your PR title suggests.

PR #1871, opened 2025-03-28
Added ApturaX

abhayporwals:


yizhao-ec:
Hello, thanks for the PR. https://github.com/abhayporwals/aptura is a Github repo URL rather than a Github organization. Given that its a single repo atm, it may make more sense to move the repo to be directly in aptos.toml instead of creating its own distinct ecosystem.

PR #1897, opened 2025-03-30
Update algorand.toml

headline-design:


yizhao-ec:
Hello thanks for the PR! I see 2 issues in this PR:

1. in order to add `HEADLINE` as a sub ecosystem for Algorand, it needs its own headline.toml file to be created
2. at the same time https://github.com/headline-design appears to be a Github user account rather than Github organization account so it should not be placed under github_organizations 

Please fix the above and then happy to merge in the PR. Thanks again.

jubos:
Thanks for the change.  Merged!

PR #1929, opened 2025-04-10
LayerZero Ecosystem Information Updates

10xKelly:
* Added sub-ecosystem projects
* Removed duplicate, outdated, and private repositories from the list
* Reorganized LayerZero Labs repositories to stay on top of the repository list
* Included additional relevant repositories from sub-ecosystem projects.


jubos:
Thanks for this.  We are in the midst of moving to the v2 data model and we can revisit this after the migration is done.  

For interop layers like Layer Zero and Axelar, we don't add ecosystems as sub ecosystems just because they have an integration or use some part of the sdk.  

The repos look reasonable to add though.

10xKelly:
> Thanks for this. We are in the midst of moving to the v2 data model and we can revisit this after the migration is done.
> 
> For interop layers like Layer Zero and Axelar, we don't add ecosystems as sub ecosystems just because they have an integration or use some part of the sdk.
> 
> The repos look reasonable to add though.

Thank you very much for your review and response. If possible, we would greatly appreciate it if you could provide additional clarity on how interoperability adoption can be reflected within your ecosystem data. Any further insight you can share would be helpful.

yizhao-ec:
Hello, sorry for the late response. We've release the new data model, let me know if you run into any issues converting. 

Regarding the interop layers conversation, currently we treat all ecosystems that integrates with 10+ chains as just multichain. The reasoning being that if an ecosystem (such as Axelar) integrates with many chains, its unfair to consider all Axelar devs as Ethereum devs since only a subset of those devs likely worked on the Ethereum integration. We are still happy to include specific repos from Axelar to Ethereum, just not the entire ecosystem.

PR #1947, opened 2025-05-03
Create 2025-05-03T021459_add_sui

cisqo:
Added Sui ecosystem, a Layer 1 blockchain


yizhao-ec:
Hello thanks for the PR, but all these ecosystems already exists. 


Using Aptos as an example, you should be able to see the current state of an ecosystem by running:

`./run.sh export -e Aptos aptos.jsonl`

PR #1953, opened 2025-05-06
Add some Stellar and Starknet repos

ofux:


JB-OnlyDust:
For more context, we are working with Onlydust.xyz which is a platform where open source projects are created by builders in the web3 space. 
We have been working with Starknet for the past 2 years and with Stellar for the past 7 months, hence the number of projects. 
Very happy to answer any question you might have !

yizhao-ec:
Hello thanks for the PR!

The Starknet repos generally looks good, however could you help me understand why a number of repos included under Stellar seem to be also Starknet? for example:

- https://github.com/Lead-Studios/veritix-web
- https://github.com/AURORALAOrg/AURORA-Frontend - a mix of both
- https://github.com/CodeGirlsInc/FrieghtFlow - description suggests stellar but readme talks about using starknet
- etc.

ofux:
> Hello thanks for the PR!
> 
> The Starknet repos generally looks good, however could you help me understand why a number of repos included under Stellar seem to be also Starknet? for example:
> 
> * https://github.com/Lead-Studios/veritix-web
> * https://github.com/AURORALAOrg/AURORA-Frontend - a mix of both
> * https://github.com/CodeGirlsInc/FrieghtFlow - description suggests stellar but readme talks about using starknet
> * etc.

Hi, thanks for pointing this out.

Some repos were indeed mistakenly mapped to those ecosystems (duplicates were mostly repos that are not specific to a single ecosystem, so I just removed them from both files). Sorry about that.

I fixed it in a new commit.

yizhao-ec:
Great thanks for the updates

PR #1961, opened 2025-05-13
Add Self Protocol

GigaHierz:
- add [Self Protocol](https://github.com/selfxyz)
- Add Self Protocol as sub ecosystem of Celo


yizhao-ec:
Hey thanks for the PR, looks like there are some small issues:

- the ecosystem names are case sensitive so please be consistent on self vs. Self
- the url in the repcon is an GitHub organization url but should be individual repo urls

GigaHierz:
hi @yizhao-ec , 

implemented requested updates. Please let me know if there is anything else missing

jubos:
Looks great, thanks!

PR #1982, opened 2025-05-30
Update Cargo.toml

Aren1414:
Optimized `Cargo.toml`: Updated dependencies, added support for `rayon` and `chrono`, improved compiler settings for better performance


jubos:
We have moved off of Rust as the core implementation, so will close this out.

PR #1985, opened 2025-06-02
Add Base ecosystem and Raize presale platform and Buster- mkt Prediction platform under Base

0xedev:


yizhao-ec:
Hello thanks for the PR. We already have Base as an ecosystem, while I'm not sure if we should create distinct ecosystems for the 2 repos you linked just yet given only association with 1 repo. 

I would recommend modifying to:

```
repadd Base https://github.com/0xedev/Raize #presale #defi
repadd Base https://github.com/0xedev/Buster-mkt #Prediction_market #defi
```

0xedev:
Thanks for the clarification, modified already.

PR #1997, opened 2025-06-15
chore: add Monad eco repositories

therealharpaljadeja:


yizhao-ec:
Hello, just as FYI, I realized too late that there were a number of urls in the original PR which were referencing GitHub organizations or users rather than specific repos. I found the corresponding repos related to Monad when I could but did end up removing a few orgs/users. See this [commit](https://github.com/electric-capital/crypto-ecosystems/commit/4180b2a3d14306e44301a4b615f4e3d8b46f80da). Please open a new PR if I missed any repos.

PR #2002, opened 2025-06-18
chore: add Noir ecosystem repositories

satyambnsal:


migrations/2025-06-18T095959_mutations
```diff
@@ -0,0 +1,6 @@
+-- add Noir ecosystem library
+ecoadd "Noir Lang"
```
emrecaliskan, line 2: Noir Lang ecosystem already exists

emrecaliskan (commented):
lgtm. Just need to remove the noir language eco addition since it already exists. Once that's done, we can merge it in. Thanks!

migrations/2025-06-18T095959_mutations
satyambnsal: Done.

PR #2012, opened 2025-06-20
Sui and Walrus updates

tedks:
Sui and Walrus ecosystem updates.


yizhao-ec:
Hello thanks for the PR!

There seem to be a large number of repos that are not public at the time of review, for example:
* https://github.com/0x0687/slotty
* https://github.com/0xobelisk/meark-website
* https://github.com/4-point-0/Kappabay
* https://github.com/Abdulazeez41/S2M
* https://github.com/abrshdir/d2c-full-stack

And there also seem to be a number of repos that may not be related to Sui, for example:
* https://github.com/0xgivn/initia-client-scaffold
* https://github.com/0xsongsu/dailytask
* https://github.com/4rjunc/nestfolio
* https://github.com/Accelar-labs/moonbeam-flow-ai
* https://github.com/aldric9/klaytn_backend

Could could help me understand why these repos were included?

tedks:
Not public repos: We accumulate repos over several months and would previously have added a visibility=False stanza in the TOML file, should we omit these entirely going forward? 

Of the repos you've linked, all except nestfolio currently have a dependency either on `@mysten/bcs` or on `@mysten/sui.js`; it looks like nestfolio was previously a Sui project or at least had the `@mysten/sui.js` dependency at an earlier point in time. 

Should I resubmit this with a more restrictive filter?

yizhao-ec:
Given the high number of non-public repos here, I would recommend removing them.

Regarding the repos with dependencies to @mysten/bcs or @mysten/sui.js, we consider importing a package to be insufficient evidence of contribution to an ecosystem. Happy to accept these repos, but do need more evidence of meaningful interactions with Sui. 

Do note that the repos identified in previous comments were just examples, there were other repos that meets one or both cases.

PR #2014, opened 2025-06-22
icp updates

letmejustputthishere:


yizhao-ec:
Hello, thanks for the PR! Overall the individual repos looks good.

However, there are some ecosystem level questions from our end:
1. Some of these new ecosystems seems too granular, eg. ElizaOS Plugins, the individual ICP organization ecosystems. If you feel strongly that some of these organizations should be connected to ICP directly, please point them out and we can handle that internally.
2. Many of these ecosystems are multichain, and therefore shouldn't be directly connected to ICP based purely on a subset of their repos relating to ICP. eg. elizaOS, Nethermind

I would request these to be removed, but happy to have a discussion on the individual level if you disagree. Thanks again.

letmejustputthishere:
hey @yizhao-ec, thanks for the review and comment!

> Some of these new ecosystems seems too granular, eg. ElizaOS Plugins, the individual ICP organization ecosystems. If you feel strongly that some of these organizations should be connected to ICP directly, please point them out and we can handle that internally.
just to make sure i understand this right: you're saying that some of the `ecoadd` shouldn't be ecosystems, but rather `repadd` to the `Internet Computer`? If so, i don't feel strongly about that, please feel free to turn them into `repadd` as you feel fit or let me know which ones exactly you'd like me to change then i'll provide the commit for that.

i decide wether something is an ecosytem based on those main questions that EC shared with us in the past:
- does the projects have a dedicated website?
- does the project have a dedicated X/ Discord account?

if you feel like this is not right, please provide an updated guideline on how to make that decision (ideally on the github, as i think this is relevant for other contributors as well)

> Many of these ecosystems are multichain, and therefore shouldn't be directly connected to ICP based purely on a subset of their repos relating to ICP. eg. elizaOS, Nethermind
- i agree, it's just difficult automating this and unfortunately we dont have the bandwidth to look into every ecosystem in detail to figure out wether it's multichain or not. if you have a pracice to do this efficiently it would be great if you could share it. are elizaOS and nethermind the only ones, or should i change others as well? 

apart from that, it's a bit difficult to decide if the project is multichain or belongs to a specific techstack / an individual blockchain based on the information shared [here](https://github.com/electric-capital/crypto-ecosystems/releases/tag/1.2). some friction
- how do i know how many chains a project supports?
  - if its a new ecosystem, i would somehow need to know the criteria for every ecosystem and apply that to the project which is not feasible
  - if the ecosystem already exists, how can i see how it has been categorized?
- its not clear that this categorisation is in the release notes, why not move it into the toplevel readme?

for the categorisation, i'm not clear what "ecosystems" exist. e.g. what ecosystem would i use to categorise a project as multichain? is it simply "Multichain"? what about evm, SVM, Move etc? i think it woulp a lot to see the different categories alongside their ecosystem names to use with the `ecocon` command.

yizhao-ec:
Sorry for the slow responses. 

Agreed that the evaluation on ecosystem - ecosystem connections is time consuming as it often requires evidence from outside of code as well, and we face similar challenges as well. 

For the time being, to get this PR merged, I would recommend remove the ecocon commands except for the ICP location specific ecosystems. 

Let me know your thoughts.

letmejustputthishere:
hey @yizhao-ec , please let me know if this is good to merge now

yizhao-ec:
Hey, I also made a few changes and removed a few ecocon as well but feel free to open a new PR if you disagree with some of the changes. Thanks again for the contributions.

PR #2019, opened 2025-06-24
Update Ronin

SM-Trung-Le:
Add games / dApp public repo to Ronin ecosystem


yizhao-ec:
Hello thanks for the PR. Could you explain why the Ragnarok Landverse repos and https://github.com/FoonieMagus/ApeironBattleDemo_IssueList should be connected to Ronin? There doesn't seem to be much evidence of Ronin usage in those repos.

SM-Trung-Le:
Hi @yizhao-ec, both those games are currently live on Ronin, source: 

https://x.com/ApeironNFT
https://x.com/ROL_Genesis (Ronin server for Landverse)

LMK if this isn't enough evidence, i can exclude them from the commit

yizhao-ec:
That makes sense, in that case I would recommend creating new ecosystem for both of those projects, connect the repos to their respective newly created ecosystems, then connect the new ecosystems to Ronin. I would like to draw the distinction between direct developer contribution to Ronin via development work to specific repos vs. deployments that we can't not see in the repo's commits.

SM-Trung-Le:
Hi @yizhao-ec i made the update to reflect the requested changes

PR #2033, opened 2025-06-27
Add Noir + Aztec repos by wamimi to ecosystem

wamimi:
This PR adds four open-source projects built with Noir and Aztec to the Electric Capital ecosystem tracker. These include zk apps and learning resources focused on zero-knowledge proofs. Repos are tagged under both the Aztec and NoirLang ecosystems.
Thanks


yizhao-ec:
Hello thanks for the PR, both Noir Lang and Aztec Protocol already exists as ecosystems in our taxonomy so please use them instead of creating new duplicated ecosystems. You can see them here:

* https://crypto-ecosystems.xyz/#noir-lang
* https://crypto-ecosystems.xyz/#aztec-protocol

wamimi:
@yizhao-ec Updated! Thanks for the clarification!

PR #2060, opened 2025-07-10
Add new Repos to Core Ecosystem

gr-akshaya:
Added repos from regional dev events in June 2025


yizhao-ec:
Hello, thanks for the PR. 

Could you update the commands such that they are: `repadd Core <repo_url>` - I believe the ecosystem's name was recently updated

Otherwise, majority of the repos looks good, only exception being: https://github.com/PuruR07/MOVIES-APP - I'm not sure if this repo is crypto related?

gr-akshaya:
Hi @yizhao-ec, Thanks for the review. I've updated the ecosystem name to Core and removed the above mentioned URL, as it wasn't related to Core or crypto (apologies — it was just frontend-related).

Appreciate your feedback!

PR #2068, opened 2025-07-15
chore: added troves.fi and endur.fi repos

hemantwasthere:
Added repos of troves.fi and endur.fi under starknet ecosystem


yizhao-ec:
Hey thanks for the PR, given both of those projects have their own github organization and sites I would recommend creating their own ecosystems and connecting those ecosystems as child ecosystem for Starknet.

hemantwasthere:
> Hey thanks for the PR, given both of those projects have their own github organization and sites I would recommend creating their own ecosystems and connecting those ecosystems as child ecosystem for Starknet.

Done!

PR #2118, opened 2025-08-21
Adding RiseIn Community Repos

Rise-In:


yizhao-ec:
Hello, thanks for the PR but looks like the ecosystem you're referring to `Educhain` does not exist. Please use ecoadd to create a new ecosystem, or find the ecosystem that most correctly represent this ecosystem in the existing taxonomy.

PR #2124, opened 2025-08-24
Add Algorand repo: Srihitha30/crypto-ecosystems

Srihitha30:
Repo: https://github.com/Srihitha30/crypto-ecosystems
Ecosystem: Algorand

- Original (non-fork) repo
- Meaningful dev commits (no bots/inflation)


yizhao-ec:
Hello, looks like this repo is not public? We can't review a PR unless we can see the repos being added.

PR #2127, opened 2025-08-29
Add repositories to Polkadot-Network

PieWol:
Adding a bunch of repositories that build on the Polkadot Network. This ranges from dApps to substrate pallets and even cryptography. 
Most of these have received a grant from the [web3 foundation](https://grants.web3.foundation/).


yizhao-ec:
Hello, thanks for the PR! I made a small update on the ecosystem name from `Polkadot-Network` to `'Polkadot Network'` to match the ecosystem names correctly.

In terms of repos, I noticed 2 repos that need further review:
* https://github.com/perun-network/polkadot-wallet-grant - seems empty and not enough content to be included
* https://github.com/element36-io/hyperfridge-r0 - I think this is better classified as `RISC Zero` rather than `Polkadot Network` but happy to discuss

PieWol:
Hey @yizhao-ec , thanks for the quick feedback. You are right. The first repo is indeed more of a research project and doesn't contain code. The hyperfridge repo is sadly not Polkadot specific as you correctly indentified. I'll remove them from the list so that we can proceed. Appreciate your help with the naming convention.

Best.

PR #2187, opened 2025-09-29
Add new tags to existing repositories

arawrdn:
#sdk #protocol #wallet


yizhao-ec:
Hey thanks again for the interest and contributions. However, the js and workflow is not the right approach in how we gather data. 

Regarding the migration changes - similar to your other PRs - we also do not plan to currently organize the taxonomy in terms of the functionality of the repos. 

Will close for now, but happy to discuss any questions or issues you have still.

PR #2195, opened 2025-09-30
added base ecosystem

MubarakNewt:


yizhao-ec:
Hey thanks for the PR. However both Base and Zora already exists:

* https://crypto-ecosystems.xyz/#base
* https://crypto-ecosystems.xyz/#zora

Will close for now, but happy to review in the future if there are any additional modifications that are net new.

PR #2304, opened 2025-10-17
Anoma: geb + goose-lean (Round 3)

mrrobotcrypto:

- geb (#research #docs)
- goose-lean (#spec #theory)

Both are public and currently not listed in the registry. Small, focused change for easier review. Cross-checked against the author's 17 previously merged PRs to avoid duplicates.


emrecaliskan (approved):
lgtm thx!

PR #2306, opened 2025-10-17
Anoma: anoma-applib + juvix-stdlib (Round 5)

mrrobotcrypto:

- anoma-applib (#sdk #library)
- juvix-stdlib (#language #stdlib)

Both are public and active. Kept the change small and focused; no duplicates with previous PRs.


emrecaliskan:
Ty for the PR! We have both of these connected to Anoma already.

PR #2313, opened 2025-10-20
Anoma: arm-risc0 + Simple-Transfer-Example (Round 7)

mrrobotcrypto:

- anoma/arm-risc0 (#zkvm #arm #rust)
- anoma/Simple-Transfer-Example (#examples #app)

Both repositories are public, active, and not currently listed in the registry. This is a small, focused change intended to ease review. Cross-checked against previously merged PRs to avoid duplicates.


emrecaliskan:
Hi, the first one already exist and is connected to Anoma. The second returns a 404. Ty!

PR #2316, opened 2025-10-20
Algorand Project DecStor added

aniketpitre:
DecStor is a modern web application that demonstrates a decentralized approach to file storage and sharing. It functions as a secure digital "wallet" where users can manage their Algorand accounts, upload files to the InterPlanetary File System (IPFS), and securely share them with other Algorand users.

The core principle of DecStor is to separate file ownership and access from the files themselves. File metadata and sharing permissions are managed by a custom backend, while the files are stored on the decentralized IPFS network. On-chain transactions on the Algorand Testnet are used to create an immutable, verifiable public record of each file share.

### DIFF
--- a/migrations/2025-05-19T235959_algorand_updates
+++ b/migrations/2025-05-19T235959_algorand_updates
@@ -64,3 +64,4 @@ repadd Algorand https://github.com/rushikesh2k04/nft-create
 repadd Algorand https://github.com/rushikesh2k04/nftdev
 repadd Algorand https://github.com/satishccy/arcraft
 repadd Algorand https://github.com/thirumaleshp/hackseries-dynamic-scan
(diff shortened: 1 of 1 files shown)


aniketpitre:
DecStor is a modern web application that demonstrates a decentralized approach to file storage and sharing. It functions as a secure digital "wallet" where users can manage their Algorand accounts, upload files to the InterPlanetary File System (IPFS), and securely share them with other Algorand users.

The core principle of DecStor is to separate file ownership and access from the files themselves. File metadata and sharing permissions are managed by a custom backend, while the files are stored on the decentralized IPFS network. On-chain transactions on the Algorand Testnet are used to create an immutable, verifiable public record of each file share.

PR #2319, opened 2025-10-21
Adding repos from London Hackathon

vestival:
Adding repos from London hackathon https://www.easya.io/events/easya-x-algorand-london-hackathon


emrecaliskan:
Thank you for the PR. LGTM!

PR #2382, opened 2025-11-03
My AlgorandProject

kranthi1306:


yizhao-ec:
Hey, thanks for the PR. However, please conform to the existing format, remove the duplicated file and the file that is just whitespace. I would also recommend rebase to the HEAD of master branch as well.

PR #2384, opened 2025-11-03
Add My Algorand Repository

Tanushree620:
Adds the Tarushree620/Algo-dapp repository to the Algorand ecosystem taxonomy.


yizhao-ec:
Hey, thanks for the PR. However seems like this repo is either private or does not exist. We generally require repos to be publicly accessible so that we can verify ecosystem relationship and fetch developer activity before including into the taxonomy.

PR #2420, opened 2025-11-21
Submitting Algorand project

adithya567-opp:


yizhao-ec:
Hey thanks for the contribution, however Algorand already exists and does not need to be added again. Also, the 2 repos being added appears to be empty or not Algorand related. Happy to discuss if you disagree.

PR #2459, opened 2025-12-11
add CELO MX ecosystem repos for EC indexing

cultureic:


yizhao-ec:
Hello thanks for the PR, can you explain to me why the new `Celo MX` ecosystem is needed? If its just to organize the repos by the geography it would be better if you tag them as Mexico instead rather than creating ecosystem per region/country.

GigaHierz:
Hi @yizhao-ec, 

the ecosystem waschanged to celo.

yizhao-ec:
Thanks for the update! A few more small issues:

*   `https://github.com/CeloMexico/CeloComposer-UPP-tandas` is empty
*   `https://github.com/motusdao/DePayments` (for Celo): The project uses general Web3 technologies (e.g., PrivyProvider, EVM-compatible chains) but lacks specific Celo references in its configuration or code.
*   `https://github.com/ReFi-Starter/regen-eliza` (for Celo): This repository is just a fork of eliza which has Celo code as built in package but I don't think this repo actually uses these plugins in a meaningful way

Please remove these repos unless you disagree with these findings.

PR #2462, opened 2025-12-12
Clean up removed repos

zhang-wenchao:
We performed an up-to-date status check on all repositories using the GitHub API.

Repositories returning a 404 status have been removed, as they are no longer accessible. This PR removes these inaccessible repositories from records.


yizhao-ec:
Hey thanks for doing this, however we plan to keep these repos in the taxonomy for now. Given that we crawl repos over time, we have commits data for many of these repos which should be correctly attributed to their ecosystems and reflected in their developer metrics. Therefore, we will keep the ecosystem - repo connections as is. 

Thanks again for the work.

PR #2540, opened 2026-01-26
submiting algorand project update

Sadiqshaik97:


yizhao-ec:
Hey are you looking to add this repo to Algorand? If so, the line should be:

`repadd algorand https://github.com/Sadiqshaik97/stellafi`

PR #2576, opened 2026-02-17
Sei-weekly-submissions-week-1

Shreyas-sei:
This PR contains submission of repositories of new applications of Sei ecosystem

### DIFF
--- a/migrations/2026-02-17T111937_Sei_update
+++ b/migrations/2026-02-17T111937_Sei_update
@@ -4,7 +4,3 @@ repadd "New Festival Town" https://github.com/AsharaStudios/nft-web
 repadd "New Festival Town" https://github.com/AsharaStudios/NewFestivalTown-IAPersonalities
 ecoadd "PredP.red"
 repadd "PredP.red" https://github.com/PredPred-HQ/PredPred-HQ
-ecoadd "Sakura Nexus"
-repadd "Sakura Nexus" https://github.com/EvanKuo2212/sakura-nexus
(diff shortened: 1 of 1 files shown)


yizhao-ec:
Hey thanks for the contribution. 

For the first 2 links, looks like they are an user and an organization but the migrations are structured to be linked to the direct repos instead. Separately, the 2 repo links you provided both seem to be not public. 

Please update the first 2 new ecosystems to be linked to their repos, and either remove the other 2 repos or modify to public repos for those ecosystems. 

We also typically want to ensure projects reach a certain size before creating their own ecosystem, so may want to verify that as well.

Shreyas-sei:
I have edited the PR with the above mentioned changes. Just a small doubt what would be a quantifiable measure for projects to have their own ecosystem, for example, new Festival Town has a good number of users: https://app.artemisanalytics.com/application-activity/sei/application/new_festival_town?cameFrom=onchainExplorer

yizhao-ec:
Thanks for the update. I think New Festive Town as an ecosystem makes sense given they are supported by a full Github organization with their own website, however I'm less confident about the others. 

For example, both of these are no longer public:
- https://github.com/EvanKuo2212/sakura-nexus
- https://github.com/tim0ps/slime-nitro-machine

Shreyas-sei:
Thank you for flagging it, I will remove the repositories that are actually private

Shreyas-sei:
PredPred-HQ actually has a website too and they are an active project building on top of Sei so I would request we keep it

PR #2720, opened 2026-03-04
Algorand project submission

lasya-563:


yizhao-ec:
Hey thanks for the PR, can you add this as a new migration file instead of modifying an existing one?

PR #2762, opened 2026-03-22
submitting algorand project india 

BDutta18:


yizhao-ec:
Hello thanks for the PR, looks like this is white space only. Closing for now.

PR #2838, opened 2026-04-20
Add AIMilestone for Conflux Hackfest 2026

NathanWu77:


yizhao-ec:
Hey thanks for the PR, but I think it would be cleaner to not create ecosystems per event and rather just keep the connection to `Conflux Network` directly. You should also quote the ecosystem name that are multi-word. 

I think the commit should be just:

`repadd "Conflux Network" https://github.com/NathanWu77/global-hackfest-2026`

PR #2870, opened 2026-05-10
new midnight ecosystem repos

laurenelee:
May 10, 2026


laurenelee:
Is there a reason why Midnight does not show up on the main page of [Developer Report](https://www.developerreport.com/)? We have an ecosystem [page](https://www.developerreport.com/ecosystems/midnight), but it would be awesome to see Midnight on the main list as well!

yizhao-ec:
We typically only show the largest ecosystems on the Developer Report website and update it monthly as ecosystems grow. We should be updating it soon and will likely include Midnight.

PR #2919, opened 2026-07-03
Adding RiseIn Community Repos: Stellar — June 2026 Builder Challenge

Rise-In:

PR #2955, opened 2026-08-06
Adding RiseIn Community Repos: Stellar — July 2026 Builder Challenge

Rise-In:


yizhao-ec:
Hello, thanks for the PR. Same question as #2919, why the separate `Stellar EC` ecosystem here?

Rise-In:
Thanks! Same as the June PR — "Stellar EC" was just an internal label on our side; these are all Stellar projects from our Rise In Builder Challenge. I've updated the migration to `repadd Stellar` and fixed the title, so they connect directly to the Stellar ecosystem. 🙏

